PR #2, opened 2023-06-06
Add an "examples" directory in lieu of "example"

gohai:
No functional changes, but easy to add a directory per model as we start developing. (The previous library had subdirectory based on the framework the example was built with - not sure if this is needed at this point.)


shiffman:
@gohai @ziyuan-linn we should probably establish a general policy re: merging pull requests. At this stage we certainly don't have to be super strict about it, but maybe each pull request should require 1 review before merging? @ziyuan-linn I'm happy for you to take the lead on merging but if you have questions or aren't sure you can tag @gohai, me, or anyone else that makes sense in?

shiffman (approved):
Looks great!

PR #3, opened 2023-06-08
Created CONTRIBUTING.md

ziyuan-linn:
Created CONTRIBUTING.md and moved the setup section from README to it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,19 @@
+## Setup
+This build uses node version 18.15.0 and npm version.
+Install nvm and run the following commands：
+```
+nvm install 18.15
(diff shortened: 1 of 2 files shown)


gohai:
I'd vote for keeping the build instructions in README, and moving the more in-depth description about the process elsewhere 🤔

shiffman:
I don't have a strong feeling one way or another but I think while we're in early stages it does probably make sense to keep the basic build instructions in the README. It might get confusing since it's a critical first step for anyone contributing but we can cross-reference or duplicate for now.

ziyuan-linn:
I think keeping the build instructions in README would make more sense for the early stages. Closing this for now.

PR #14, opened 2023-06-26
Serve example directory via webpack devServer

gohai:
Many ml5 examples need a running webserver (rather than the file system) to function correctly. This change exposes the "examples" directory as "/examples", and the built library as "/dist", when running "npm run start". Additionally, webpack is prompted to automatically open a browser with the directory listing of the example directory when being run.

The old library appears to have had a separate web server, running on a different port, to host the example files. I believe doing it this way (and keeping the directory structure the same between http and the file system) is slightly more easier to comprehend.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,9 +25,9 @@ You should see something similar to this in the terminal:
 webpack 5.76.1 compiled successfully in 8360 ms
 ```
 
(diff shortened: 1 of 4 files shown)


ziyuan-linn (approved):
I tested it and it works great for me! I think this change would make things easier.

ziyuan-linn:
I edited the setup guide in README.md and CONTRIBUTING.md to match this new change.

shiffman (commented):
Yes, this is fantastic! Merging!

PR #15, opened 2023-07-03
model handpose

ziyuan-linn:
- Implement the palm and hand landmark detection feature.
- Upgrade the old model to the MediaPipe hand-pose-detection model.
- Add a p5 example that demonstrates this feature.
- Fix a bug where `ml5.handpose()` does not return the model correctly when the callback is not supplied
- Bring over the test file from the old library (untested)

This new model now supports 3D hand landmarks and tracks multiple hands simultaneously! A handpose example is also added to the example directory. Please let me know if I should change anything here or work on anything else.

### DIFF
--- a/examples/Handpose/index.html
+++ b/examples/Handpose/index.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Document</title>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.6.0/p5.js"></script>
+    <script src="../../dist/ml5.js"></script>
+  </head>
+  <body>
+    <script src="script.js"></script>
+  </body>
+</html>
--- a/examples/Handpose/index.html
+++ b/examples/Handpose/index.html
@@ -6,7 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Document</title>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.6.0/p5.js"></script>
-    <script src="../../dist/ml5.js"></script>
+    <script src="http://localhost:8080/ml5.js"></script>
   </head>
   <body>
     <script src="script.js"></script>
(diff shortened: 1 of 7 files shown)


yining1023 (dismissed):
LGTM!! This is amazing, @ziyuan-linn! The model and the example work so well, great grest job!! 👏 👏 👏 

Left some minor and unblocking comments inline. I noticed a small thing where the `flipHorizontal` might need to be wrapped in `{}`. Feel free to merge it after solving the `flipHorizontal` issue!

For the next steps, I'm thinking:
1. Maybe we can talk about the output format with the rest of the team in the next meeting. I think it would be nice to expose some keypoints("wrist", "thumb", "indexFinger"...) by name to users, like we did in [posenet](https://learn.ml5js.org/#/reference/posenet?id=singlepose)
```
[
    {
        "handedness": "Left",
        "wrist": { <-----------------------add
            "x": 57.2648811340332,
            "y": 489.2754936218262,
            "z": 0.00608062744140625,
            "confidence": 0.89
        },
        "thumb": [ <-----------------------add
            {
                "x": 57.2648811340332,
                "y": 489.2754936218262,
                "z": 0.00608062744140625,
                "confidence": 0.89,
            },
            {...},
            {...},
            {...},
            {...},
        ],
        "indexFinger":[], <-----------------------add
        "middleFinger":[], <-----------------------add
        "ringFinger":[], <-----------------------add
        "pinky":[], <-----------------------add
        "keypoints": [
            {
                "x": 57.2648811340332,
                "y": 489.2754936218262,
                "name": "wrist"
            },
        ],
        "keypoints3D": [
            {
                "x": -0.03214273601770401,
                "y": 0.08357296139001846,
                "z": 0.00608062744140625,
                "name": "wrist"
            }
        ],
        "score": 0.9638671875,
    },
    {
       "handedness": "Right",
       ...
    }
]
```
Alternatively, we can also add each point in each finger as well. [More info](https://github.com/tensorflow/tfjs-models/blob/master/hand-pose-detection/src/constants.ts).
```
[
  'wrist',
  'thumb_cmc',
  'thumb_mcp',
  'thumb_ip',
  'thumb_tip',
  ...
]
```

2. We can bring over the jest test [script](https://github.com/ml5js/ml5-library/blob/f80f95aa6b31191ab7e79ff466e8506f5e48a172/package.json#L20C3-L20C25) since this PR adds test now.
3. Sth we can talk about in the meeting is that I noticed that `handedness` definition was a little confusing to me. I thought it can detect if it was left or right hand, but the results were always the opposite to me. Maybe it means the hand on the left side of the screen?

ziyuan-linn:
@yining1023 Thank you again for your review! I made changes to the example file as you suggested. I also added filtering to the option arguments for both initialization and inference so we only pass in the options needed. 

I will merge this now! I think the next step would be adding comments about the options in the example and wrapping the prediction output in a more user-friendly format after talking about it with the team next meeting.

Issue #18, opened 2023-07-14
NPM vs Yarn


ziyuan-linn:
Hi everyone,

I am trying to install some packages for MoveNet with npm and got `ERESOLVE unable to resolve dependency tree` error. 

<img width="634" alt="Screenshot 2023-07-14 184943" src="https://github.com/ml5js/ml5-next-gen/assets/93690311/d0c56da9-f735-4710-aa24-f90a17b27f0c">

It seems like TensorFlow is using dependencies with conflicting versions and npm does not support it. 

The TensorFlow [installation guides](https://github.com/tensorflow/tfjs-models/tree/master/pose-detection/src/movenet#installation) for newer models use Yarn instead of npm when installing the packages. I tried installing MoveNet packages with Yarn instead of npm and was successful. I think the task of managing dependencies will be easier with Yarn, and I'm wondering whether we should switch our package manager.

PR #19, opened 2023-07-17
change package manager

ziyuan-linn:
- switch package manager from npm to yarn
- update lock file
- update documentation to use yarn instead of npm

This PR should resolve the [version conflict issue](https://github.com/ml5js/ml5-next-gen/issues/18) when installing dependencies with npm.


ziyuan-linn:
@shiffman @yining1023 I added a brief explanation as to why we are switching to yarn. Please let me know if I should change anything else!

shiffman:

PR #21, opened 2023-07-23
Simplify Handpose Example

shiffman:
I've made some simplifications to the handpose example. These apply here, but also more broadly to how we will build out the examples for the rest of the library. These are the changes I made to match p5.js style.

1. Removed the use of `=>`. 
2. Broke out callbacks into named functions rather than anonymous.
3. Removed the use of `const`.
4. Renamed JS file from `script.js` to `sketch.js`

We could choose to break with these conventions in our examples, but maybe we would want to include some material on the website for people who are coming from p5.js?

Related discussion: #22.


ziyuan-linn (approved):
I think these changes are great!! This is definitely easier for people familiar with p5.js or new to javascript.

shiffman:
@ziyuan-linn I realize there may be some differences with these changes and the discussion in #22. Shall we merge this pull request now and then you can adjust as necessary?

ziyuan-linn:
@shiffman Yes, I think we can merge this!

PR #26, opened 2023-08-01
Add a LanguageModel class that implements the Llama2 architecture via llama2.c and Emscripten

gohai:
This makes it possible to do inference on toy-models (15M, 42M, 110M parameters) using the Llama2 architecture, as implemented in [llama2.c](https://github.com/karpathy/llama2.c/).

The included emscripten build artifacts came from this tree: https://github.com/gohai/llama2.c-emscripten

TODO:
- model is hot-linked against Hugging Face's server - not sure if this is proper, or if this should live on a CDN instead?
- this currently tries to accommodate every flavor of writing JS (promise, callback, event) - possibly limit the options?
- is there a way to have p5's `preload` wait for an arbitrary promise to resolve?
- not firm about naming of things ("LanguageModel"?, "manual*"?, "tokens"?)
- would be nice to come up with a Colab notebook to do training/fine-tuning of custom models?


examples/LanguageModel/sketch.js
```diff
@@ -0,0 +1,61 @@
+let lm;
+let curWord = 0;
+let angle = 0;
+let radius = 60;
+
+function setup() {
+  createCanvas(400, 400);
+  background(0);
+
+  lm = ml5.languageModel(onModelLoaded);
```
shiffman, line 10: Agreed it would be great to support `preload()`. 

Also, I wonder if we should require a "string" that references a specific model to load. Even though this is perhaps an extra, unecessary step it emphasizes to the user that this isn't magic, and requires an acknowledgement of the specific model they are using. Some options:

```javascript
  lm = ml5.languageModel("TinyStories", onModelLoaded);
  lm = ml5.languageModel("TinyLlamas", onModelLoaded);
```

Referencing the model name is probably more important, but it's also important to emphasize the specific dataset that was used for training.

examples/LanguageModel/sketch.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+function generateText() {
+  let prompt = select('#prompt').value();
+  console.log('Prompt: ' + prompt);
+
+  let options = {
+    temperature: 0.9
+  };
+  lm.generate(prompt, options, onToken);
```
shiffman, line 31: I love the idea of an `onToken()` event, but for beginners it would be incredibly useful to have an API that mirrors the HuggingFace inference API. In this case, you pass a prompt and a desired number of tokens and you get the entire result back. Something along the lines of:

```javascript
  let options = {
    temperature: 0.9,
    maxTokens: 100
  };
  lm.generate(prompt, options, gotText);
```

I might also consider just be the form `prompt, maxTokens, callback` and then a JS object with more properties if you wanted to set temperature and other more, maybe:

```javascript
  lm.generate(prompt, 100, gotText);
```

and then:

```javascript
  let options = {
    prompt: "How are you?",
    temperature: 0.9,
    maxTokens: 100
  };
  lm.generate(options, gotText);
```

Though maybe it's good to always have the prompt separatel

shiffman (commented):
Wow, this is amazing @gohai! 

First, as we work on documenting this, a wonderful reference is [Let Us Show You How GPT Works](https://www.nytimes.com/interactive/2023/04/26/upshot/gpt-from-scratch.html) by @aatishb. Hi Aatish, tagging you mostly to say hi, but would of course welcome your input!

I added a few comments @gohai. In my experience with teaching the previous `charRNN` models with ml5.js, the main pain points were:

* Generating character by character or token by token was very confusing for students and should probably be reserved for more advanced use cases (though great to include in ml5.js if we can?)
* What students often want to do is train their own model, but I was never able to successfully pull this off with all the steps required to train in python then convert to JS. A clear and easy to use colab notebook could absolutely work though!

My other "worry" relates to treading into this territory and providing "out of the box" models. LLMs, as is well documented, have all sorts of bias and other issues. It would wonderful to collaborate with the team members who are working on educational materials and documentation to think about how we guide users to be aware of the limitations and possible dangers of certain models and datasets.  The code of conduct can be a good reference as well for considering use cases and applications. This could be a great discussion for the full group! cc @sproutleaf

gohai:
@aatishb I so love this article, particularly how it raises the transparency/blackbox-ism issues with OpenAI to general audience! (Believe that's also at the core for why bother with toy models - not to suggest a similar type of performance is achievable in the browser, but to unpack this technology, and how it got built/trained, as widely as possibly.)

**Thank you** for taking the time to think through this, @shiffman!

- The suggestion of requiring the user to spell out the model name makes so much sense! (implemented this in https://github.com/ml5js/ml5-next-gen/commit/4eb89d9a4ad111c22cddd7e6e7dd528c421618eb) I feel similarly about what you wrote about the dangers of whatever the "out-of-the-box" experience is going to be. Looking forward to working with the larger group on this!
- Following your suggestion, I changed the main callback to be on-completion rather than on-token and simplified the most basic example. (https://github.com/ml5js/ml5-next-gen/commit/b544b2432d5873987d6687733a0eb93d6eb313d9) 
- I'll look into training of custom models. From scratch on the TinyStories dataset seems to take a "couple of hours" on four A100. This is probably a step to far for Colab 😄, but perhaps there will be improvements in the implementation. There are people [working](https://github.com/karpathy/llama2.c/pull/187) on implementing fine-tuning using LoRA, this might be doable!
- And yes to preload!

examples/LanguageModel/sketch.js
```diff
@@ -0,0 +1,50 @@
...
+  let prompt = select('#prompt').value();
+  console.log('Prompt is "' + prompt + '"');
+
+  let options = {
+    temperature: 0.9
+  };
+  lm.generate(prompt, options, gotText);
+}
+
+function gotText(out, lm) {
```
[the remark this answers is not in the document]
gohai: @shiffman `Words` and `out` are properties of the instance currently - would passing the instance as the result make sense, or would you feel a "result" object value is nicer, which the user is able to mutate as they like etc? (`text` over `out` for the literal output?) I'll add `prompt`, although currently its strictly the beginning of the response - more of a "prefix" really.
Thank you for your review! I'll try it out

PR #23, opened 2023-07-23
Basic Neural Network Examples

shiffman:
I'm working on [Chapter 10 of the Nature of Code](https://nature-of-code-2nd-edition.netlify.app/neural-networks/). The goal of the chapter is to demonstrate neuro-evolution with ml5.js for autonomous steering behaviors. However, before I get to that I would like to demonstrate the basics of training a neural network classifier with ml5.js. The chapter outline goes something like:

1. What is a Neural Network?
2. Perceptron (p5.js only)
3. Multi-Layered Perception with ml5.js
    * Color Classifier
    * Other examples?
4. Neuro-Evolution with ml5.js
    * Simple Steering
    * Simple game? (chrome dinosaur jumping?)
    
 I'm working on the examples here in this branch and would welcome feedback! I'm especially struggling with a good "hello world" neural network example. It would be great to demonstrate something like hand pose classification, but that is probably beyond the scope of what I can do with the chapter. Right now, this "color classifier" is my best idea.

Feedback welcome!

PR #28, opened 2023-08-06
Add the SoundClassifier model

ch3926:
This PR brings soundClassifier into the new ml5 library. The code from the old ml5 library was largely unchanged except for minor issues caused by recent Webpack updates.

I have conducted the following tests:

1. Tested example code from old repo 'SoundClassification_speechcommand_load' --> uses pre-trained model ‘SpeechCommands18w’ which recognizes 18 sounds.

2. and 'SoundClassification_speechcommand'. --> uses a custom model I trained with Teachable Machine to recognize clapping and whistling sounds.

Feedback welcome!


examples/SoundClassification/sketch.js
```diff
@@ -0,0 +1,47 @@
...
+
+    document.body.appendChild(label);
+    document.body.appendChild(confidence);
+    // Classify the sound from microphone in real time
+    classifier.classify(gotResult);
+}
+setup();
+
+// A function to run when we get any errors and the results
+function gotResult(error, results) {
```
shiffman, line 37: Noting that in #23 I changed the pattern of "error first" callbacks and this should now work with the simplified:

```javascript
function gotResult(results) {
}
```

Error checking is now optional!

```javascript
function gotResult(results, error) {
}
```

shiffman (commented):
Great to see this coming through! Made a comment about the order of arguments in callbacks that changed in #23.

examples/SoundClassification/sketch.js
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2019 ml5
```
gohai, line 1: If existing examples written using p5.js work similarly well to this (vanilla-JS) one, I'd suggest using the p5.js-based ones (without async/await) instead.

src/SoundClassifier/index.js
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) 2019 ml5
```
gohai, line 1: Nitpick: whenever you're making the slightest modification of a file, you'd be entitled to update the copyright header to e.g. `2019-2023` 🙂

src/SoundClassifier/index.js
```diff
@@ -0,0 +1,90 @@
...
+        return this.classifyInternal(numberOfClasses, args.callback);
+    }
+}
+
+const soundClassifier = (modelName, optionsOrCallback, cb) => {
+    const { string, options = {}, callback } = handleArguments(modelName, optionsOrCallback, cb)
+        .require('string', 'Please specify a model to use. E.g: "SpeechCommands18w"');
+
+    let model = string;
+    // TODO: I think we should delete this. -Linda
```
gohai, line 81: Since this model is getting some testing presently, now is probably a good time to follow Linda's suggestion here, and see if anything breaks!

examples/SoundClassification/index.html
```diff
@@ -0,0 +1,21 @@
...
+
+<body>
+    <h1>Sound classification using SpeechCommands18w</h1>
+    <p>Try to speak the following commands to your microphone: 'zero', 'one', 'two', 'three', 'four', 'five', 'six',
+        'seven', 'eight', 'nine', 'up', 'down', 'left', 'right', 'go', 'stop', 'yes', and 'no', in addition to
+        'background_noise' and 'unknown'</p>
+    <script src="sketch.js"></script>
+</body>
+
+</html>
```
gohai, line 21: FYI: this symbol (and the associated warning you might see when committing) comes from Git expecting that there's a new-line character at the end of each file.

Our repo has `prettier`, which can take care of such minor issues. You could run `npm run format` and then just commit the changes pertaining to your code to the branch.

gohai (commented):
@ch3926 👏 I left some comments. Let's talk if anything doesn't make sense (which might well be the case :)

ch3926:
Thank you both for the feedback. I will look into these fixes asap and let
you know if I have any questions!

Best,
Connie

examples/SoundClassification/index.html
```diff
@@ -0,0 +1,21 @@
+<html>
+
+<head>
+    <meta charset="UTF-8">
+    <title>Sound classification using SpeechCommands18w</title>
+    <script>
+        // Required by some npm packages
+        window.process = { browser: true, env: { ENVIRONMENT: 'BROWSER' } };
```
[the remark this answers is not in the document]
ch3926: I tested the code and it seems to work fine without it. Perhaps the problem has been solved since 2019!

src/SoundClassifier/index.js
ch3926: I tried running with it deleted and got an error message "Unsupported URL scheme in metadata URL:..."

These convert 'model' from **'SpeechCommands18w' to 'speechcommands18w'**
So I tried deleting the lines again but also changing the initializer line in the example code from **uppercase to lowercase** 'classifier = await ml5.soundClassifier("**speechcommands18w**", options);' This works.

I wonder what she meant by delete. Perhaps, she means to consider whether there is a better solution?

gohai: Thank you for figuring out what this was for, @ch3926!

And yes, I think we should keep it: the library assumes all lower-case, but existing code [seems to use](https://learn.ml5js.org/#/reference/sound-classifier?id=quickstart) camel-case. Perhaps better solving this in the constructor, though, with a helpful comment.

lindapaiste: @ch3926 I think I was interpreting the `if` statement backwards when I left that comment! 🙃 We do not want to modify the casing if the user provides a URL.  So you can delete my comment.

shiffman:
Hi @ch3926, at a recent meeting @OrpheasK expressed interest in revisiting this branch and finishing off any last details! If you have any questions or specific concerns let us know, otherwise we hope to be able to add to it and merge soon!

ch3926:
That sounds great! I'm happy to see this branch progress!

Best,
Connie

OrpheasK:
This bring the Sound Classifier model up to date with the rest of the library. There are no significant alterations to the contributions @ch3926 has made other than reordering the callback arguments and incorporating the new `getTopKClasses` based on @lindapaiste's previous [PR](https://github.com/ml5js/ml5-library/pull/1364). The SoundClassification examples are made into p5 sketches that are somewhat beginner-friendly but can also be further simplified

shiffman:
This is fantastic @OrpheasK, so glad that @ch3926's work will be merged soon! Unfortunately a small conflict was introduced via some of the other pull requests I merged, could you take a look?

@lindapaiste let us know if you have any feedback regarding incorporating the new `getTopKClasses` and your work with `preload`.

Here's a question! Shall we change `classify()` to `startClassify()`?!?!? This would align a bit more with the API for image classification and the pose detection models.  Also, in looking at the examples, I find it a little odd that the microphone input is not specified anywhere.  Certainly it's nice that it would use the mic by default, but I wonder if we should tie this into `p5.sound` more explicitly and have the examples use [`p5.AudioIn`](https://p5js.org/reference/#/p5.AudioIn) or at least have some kind of constant or options object passed in via `startClassify()`. Could this actually work with a sound file or other audio input.

I don't mean to open a big can of worms. I'm also ok with merging as is and discussing this in a separate issue?

Welcome any thoughts!

src/SoundClassifier/speechcommands.js
```diff
@@ -0,0 +1,52 @@
...
+      .listen(async (result) => {
+        if (result.scores) {
+          const classes = await getTopKClasses(
+            result.scores,
+            topk,
+            this.allLabels
+          );
+          return cb(classes);
+        }
+        return cb(`ERROR: Cannot find scores in result: ${result}`);
```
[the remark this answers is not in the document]
OrpheasK: Changed it just in case!

OrpheasK:
Using `classifyStart()` for both the Image and Sound Classifiers makes sense to me, haven't really into using p5.sound/audio files with this model but definitely up to integrating something along those lines! Either in this or a separate PR

shiffman:
> Using `classifyStart()` for both the Image and Sound Classifiers makes sense to me, haven't really into using p5.sound/audio files with this model but definitely up to integrating something along those lines! Either in this or a separate PR

How about we switch to `classifyStart()` in this branch before I merge? (Could you add a `classifyStop()` as well and then open a separate issue to discuss future support for custom audio inputs as well as a one-time `classify()` method for a short sound clip?

OrpheasK:
Sounds good, getting on to this!

OrpheasK:
It seems an error occurs after merging main that has to do with the way the Sound Classifier is loaded through p5Utils/preload, (I was using the now deprecated p5PreloadHelper when developing this). Will look into this shortly.

shiffman:
Sounds good, thank you @OrpheasK! Let me know when this is ready for review again!

OrpheasK:
Model is now loading properly thanks to @ziyuan-linn's fix which means this is ready for review @shiffman!

shiffman (approved):
Yaaaay!!!! What a long journey this one has been! I made a couple tiny changes for style. @ziyuan-linn perhaps you can add these examples now to the web editor collection.

Here we go!!!!!!

PR #32, opened 2023-08-13
format all js documents

ziyuan-linn:
- run the `yarn run format` command
- formatted all .js documents according to the Prettier rules

@shiffman This PR should fix the issue mentioned in #31 :)


shiffman (approved):
Thank you @ziyuan-linn!

Issue #22, opened 2023-07-23
API for continuous vs. one-time body/hand/face detection


shiffman:
I'm opening this thread to discuss aligning the API across all keypoint detection models (currently planned: hand, body, face). This relates to #21.

The hand model implemented by @ziyuan-linn uses an event callback with `handpose.on("hand", gotHands);` I like this pattern and it would make sense to adopt it for body pose and face keypoint tracking. However, should we also provide a function for a "one-time" hand detection in an image? Here's one idea:

### Continuous
```javascript
// continuous if a video element is passed into handpose() and on() is used? 
// Or should video be passed into on()?
let handpose = ml5.handpose(video, modelReady);
handpose.on("hand", gotHands);
function modelReady() {
  console.log("model ready");
}
function gotHands(results) {
  console.log(results);
}
```

### One-time
```javascript
// No media is passed in when loading the model
let handpose = ml5.handpose(modelReady);
function modelReady() {
  console.log("model ready");
  // Individual image passed in?
  handpose.detect(img, gotResults);
}
function gotHands(results) {
  console.log(results);
}
```

This brings up two other key questions:

### Support asnyc/await?

Should we support `async` and `await` or keep things simple and more aligned with p5.js callbacks only? The following syntax is so lovely, but not all how p5.js does things! cc @Qianqianye

```javascript
async function runHandPose() {
  let handpose = await ml5.handpose();
  let results = await handpose.detect(img);
  console.log(results);
}
```

### Error first callbacks?

The p5 style for callbacks is to have one object containing the results.

```javascript
function gotHands(results) {
  console.log(results);
}
```

ml5.js has had "error first" callbacks for most of the models in prior versions:

```javascript
function gotHands(error, results) {
  console.log(results);
}
```

My preference for beginners is the p5.js style (and include an optional second `error` argument for debugging). However, I believe ml5.js adopted the error first callback previously because (a) it is a JS convention and (b) it was required for us to be able to support both callbacks and async/await. @joeyklee tagging you in here, is my memory of this correct?

I might lean towards keeping things simple (a) not support asnyc/await (radical!) and therefore (b) not require error-first callbacks. But I'm not at all sure here, that async/await syntax is quite nice!

@MOQN, would love your thoughts as someone who has taught these models quite a bit.

limzykenneth:
Really interesting ideas here, thanks @Qianqianye for pointing me to this issue.

[...]

PR #35, opened 2023-08-16
Handpose interface

ziyuan-linn:
Hi everyone, I implemented a new handpose API according to this discussion #22. 
- add `detect()`, `detectStart()`, and `detectStop()` function
- remove EventEmitter interface
- add support for p5 preload to `ml5.handpose()` function
- add examples for single image detection and start/stop functionality

Some issues that came up during implementation: 

The current `ml5.handpose` function returns an `Handpose` object in order to support p5 preload. Thus it currently cannot support the promise interface. There might be a workaround to support both interfaces, but I will reach out to the p5 team and get a better grasp of `preload()` first.

If `detectStart()` is called twice, for instance:
```javascript
detectStart(video, gotHands);
detectStart(video2, gotHands);
```
The model would not output any warning and output detection for video2 only. Please let me know if this behavior makes sense.

To-Dos:
- figure out if both p5 preload and promise interface can be supported
- implement named landmark output


examples/Handpose-p5-image/sketch.js
```diff
@@ -0,0 +1,39 @@
...
+  handpose = ml5.handpose();
+}
+
+function setup() {
+  createCanvas(640, 480);
+  fill(0, 255, 0);
+  noStroke();
+  // Draw the image
+  image(img, 0, 0);
+  // start detecting hands from the webcam video
```
shiffman, line 23:
```suggestion
  // Detect hands in an image
```

examples/Handpose-p5-image/sketch.js
```diff
@@ -0,0 +1,39 @@
...
+  handpose.detect(img, gotHands);
+}
+
+// Callback function for when handpose outputs data
+function gotHands(results) {
+  // save the output to the hands variable
+  hands = results;
```
shiffman, line 30: For this example, if the drawing is happening in `gotHands()` I don't think we need a global variable. I would either just draw from `results` only or move drawing to the `draw()` loop? Not sure which is a better foundational example? You could also name the argument to `gotHands()` as `hands` if you like?

examples/Handpose-p5-start-stop/sketch.js
```diff
@@ -0,0 +1,69 @@
...
+
+let handpose;
+let video;
+let hands = [];
+let isDetecting = false;
+
+function preload() {
+  // Load the handpose model.
+  handpose = ml5.handpose();
+  console.log("preload");
```
shiffman, line 14: I would remove this or change it to "Hand Pose model is loaded." or something like that?

examples/Handpose-p5-start-stop/sketch.js
```diff
@@ -0,0 +1,69 @@
...
+}
+
+function setup() {
+  createCanvas(640, 480);
+  fill(0, 255, 0);
+  textSize(25);
+  noStroke();
```
shiffman, line 21: I think it's generally more helpful and clear to place these where the drawing is happening (even though it's redundant). With these in `setup()`  I find it can cause confusion for students adapting the examples to include more elaborate drawing.

shiffman (commented):
This is wonderful @ziyuan-linn! A few thoughts below:

1. If `startDetect()` does not support more than one video source (which is fine), we should make this clear in the documentation (cc @MOQN and @sproutleaf) and maybe output a warning if it's called twice.

2. It would be nice to include an example that shows how to to track maybe 2-3 points from the hand by name. This is often what I start people with. One idea would be to check the distance between the top point of the index (pointer?) finger and top point of the thumb. That could be mapped to the size of a circle for example? I'm happy to work on this if you like!

Other comments include inline with the code!

examples/Handpose-p5-image/sketch.js
ziyuan-linn: Committing this suggestion!

examples/Handpose-p5-image/sketch.js
ziyuan-linn: Thank you for this suggestion! I would lean toward moving the drawing to the `draw()` loop since the students might want to modify the example into something more interesting and it would be easier if the drawing is already in `draw()`.

examples/Handpose-p5-start-stop/sketch.js
ziyuan-linn: Ah, this was for testing and I totally forgot to remove it. Thank you for pointing it out, removing now!

examples/Handpose-p5-start-stop/sketch.js
ziyuan-linn: I have moved these to their corresponding locations!

ziyuan-linn:
@shiffman thank you for your review! I made some changes according to your inline suggestions. I added a warning when `detectStart()` is called twice without a call to `detectStop()` in between. I also added named keypoints (named after the tf.js' `name` field for each landmark point). 

Please let me know if there are any other issues that need to be addressed. Otherwise, I think we can merge this so you can add the keypoint-by-name examples?

shiffman:
Looks great @ziyuan-linn! I'm going to go ahead and merge. If others have comments or questions, we can address them in separate issues / branches!

PR #36, opened 2023-08-19
Adding ml5.setBackend()

shiffman:
@ziyuan-linn, what are your thoughts on the best way to add this method?


ziyuan-linn:
@shiffman Ah, I did not know this branch existed. My implementation is here: https://github.com/ml5js/ml5-next-gen/tree/ml5-set-backend. Sorry about putting things in two places.

I put `setBackend()` in one of the util files and added a custom error since the tf.js is not very intuitive. I think we also have to somehow make it clear this method will only work for tf.js-based models. Documenting it on the website might be a good idea.

shiffman:
@ziyuan-linn did you push your commits to that branch? I just merged them here but don't see them now! It's probably best for you to open a new pull request and I'll just close this one!

ziyuan-linn:
@shiffman Ah yeah I forgot to push my commits. I will open a new pull request!

shiffman:
Great! (Closing and moving to #37)

PR #37, opened 2023-08-20
add setbackend

ziyuan-linn:
Add the `ml5.setBackend()` method


shiffman (approved):

shiffman:
Awesome! Will merge and adapt my neuroevolution examples to work with this new method!

PR #43, opened 2023-09-05
Update Sentiment Analysis Model to version 2.0

ch3926:
This version is compatible with the main branch and contains a p5 example.

Main Updates:
- Src is largely unchanged from previous version
- Imported modelLoader
- Updated p5 sketch with new functionality
- Should we pick a more sophisticated model? Currently uses this [movie dataset.](https://keras.io/api/datasets/imdb/).


ch3926:
@gohai @shiffman This is the new PR for sentiment-analysis. It is currently ready to be merged into main.

gohai:
Thank you, @ch3926! 🙌 I tested your branch and committed the few odd cleanups.

@shiffman That'd be ready to be merged if we still plan on keeping the sentiment analysis model around.

shiffman:
Wonderful, great work @ch3926! Merging! This is a lovely demonstration and example for my class this fall. Tagging in @sproutleaf @QuinnHe @MOQN just to keep you in the loop in terms of documentation for the new website! 

@gohai If we end up building more features into ml5.js as they relate to transformer models, I could see us replacing this with a different method, we can continue to discuss!

PR #46, opened 2023-09-24
add production build config

ziyuan-linn:
This PR refactors the WebPack config file and adds a production build option.
Run `yarn run build` to create a production build of the library in the `dist` folder.


shiffman:
Looks great @ziyuan-linn, thank you!

PR #47, opened 2023-10-04
add webpack-merge

ziyuan-linn:
add the webpack-merge dependency

It is used by webpack.config.js and I forgot to add it.


shiffman:
Great!

PR #48, opened 2023-10-04
Add blazepose

ziyuan-linn:
This PR adds the BlazePose model to BodyPose. 
The changes are as follows:
- add a third parameter, modelName, to `ml5.bodypose` which specifics the underlying model to use. It defaults to MoveNet if no model name is provided.
```javascript
ml5.bodypose("BlazePose", options, callback);
//or 
ml5.bodypose("MoveNet", options, callback);
```
- update config filtering based on which underlying model is being used. 
- update `addKeypoints` function so it adds the z values to the named keypoints (when using BlazePose).
- add a simple example for using bodypose in BlazePose mode.

### DIFF
--- a/examples/Bodypose-blazepose-keypoints/index.html
+++ b/examples/Bodypose-blazepose-keypoints/index.html
@@ -1,0 +1,19 @@
+<!--
+ Copyright (c) 2018-2023 ml5
+ 
+ This software is released under the MIT License.
+ https://opensource.org/licenses/MIT
+-->
+
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <title>ml5.js Bodypose Blazepose Detection Example</title>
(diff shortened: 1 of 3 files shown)


ziyuan-linn:
Hi @shiffman, thank you for the review! I have added a warning for when the user uses an unknown model name.

I think we could implement a system in the near future that checks any user input and gives friendly errors accordingly. The rough idea is:

```javascript
// checkString(userInput, stringsToMatch, ErrorMessage)
// matches userInput with any element in stringsToMatch, case-insensitively.
// @returns the lowercase version of userInput if valid, otherwise returns the first element of stringsToMatch and log the error.
this.modelName = checkString(this.modelName ["movenet", "blazepose"], "Error: unknown model name...using default.");
```

shiffman:

PR #52, opened 2023-10-11
fix p5 reserve word error

ziyuan-linn:
Currently p5.js always outputs a warning that `you have used a p5.js reserved variable "VERSION"` when ml5 is present. See the screenshot below.

Screenshot 2023-10-11 175143<br/>
I managed to narrow down the problem and it is not in our code. It seems like one of the dependency files for facemesh, `node-modules/@tensorflow-models/face-landmarks-detection/dist/face-landmarks-detection.esm.js`,  declares a global variable "VERSION" for some reason (the code is minified so it is hard to tell). After removing the global variable declaration from the dependency file, p5 no longer outputs the error, and it does not seem to affect facemesh in any way. This PR adds a script to automatically remove the piece of code causing the trouble after each dependency install. 

Changes:
- install patch-package tools
- modify `package.json` to stop automatic update of facemesh's dependency
- add a patch file that specifies a change to the content of facemesh's dependency
- add an npm script that applies the change automatically to  facemesh's dependency after each install

This should temporarily work but it is somewhat of a hack. Perhaps we should eventually reach out the the tf.js team about the issue and see if the global declaration can be changed?


shiffman:
We're discussing at our meeting. I am going to merge this, but @ziyuan-linn when you have a chance open an issue as a reminder to maybe look into a better, more permanent solution.

PR #54, opened 2023-10-12
update package name

ziyuan-linn:


shiffman:
2nd time is the charm!

PR #59, opened 2023-10-18
install rimraf

ziyuan-linn:
rimraf is used to remove the dist folder before the build process, but it was never installed.

Changes:
- install rimraf to dev dependencies


shiffman:
Looks good, thanks for catching this @ziyuan-linn!

PR #45, opened 2023-09-16
Facemesh: Add bounding boxes to each named keypoint in result

gohai:
This adds a bounding box to each named part of the face - as discussed [here](https://github.com/ml5js/ml5-next-gen/pull/42#discussion_r1317387553).

PR #64, opened 2023-11-06
fix centerY typo

ziyuan-linn:
Looks like the fix somehow did not get through when we merged #45. Applying the fix again in the PR.


shiffman:
Apologies if I missed something, merging now!

PR #66, opened 2023-11-12
Improve comments in example sketches

ziyuan-linn:
This PR has no functional changes and mostly modifies comments in the example sketches.
- Format the inline comments in the example sketches to keep them consistent
- Fix a few typos in the comments of BodyPose's examples
- Rephrase a few comments in Sentiment's example
- Add copyright and license headers to the documents that didn't have them

### DIFF
--- a/examples/Handpose-keypoints/index.html
+++ b/examples/Handpose-keypoints/index.html
@@ -1,3 +1,10 @@
+<!--
+ Copyright (c) 22023 ml5
+ 
+ This software is released under the MIT License.
+ https://opensource.org/licenses/MIT
+-->
+
 <!DOCTYPE html>
 <html lang="en">
   <head>
--- a/examples/Handpose-keypoints/index.html
+++ b/examples/Handpose-keypoints/index.html
@@ -1,5 +1,5 @@
 <!--
- Copyright (c) 22023 ml5
+ Copyright (c) 2023 ml5
  
  This software is released under the MIT License.
  https://opensource.org/licenses/MIT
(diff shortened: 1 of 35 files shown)


examples/Handpose-keypoints/index.html
```diff
@@ -1,3 +1,10 @@
+<!--
+ Copyright (c) 22023 ml5
```
shiffman, line 2:
```suggestion
 Copyright (c) 2023 ml5
```

examples/NeuralNetwork-color-classifier/sketch.js
```diff
@@ -1,3 +1,8 @@
+// Copyright (c) 2023 ml5
```
shiffman, line 1: This is maybe a silly comment, but I sometimes find that when looking at the examples that the copyright notice at the top is a little off-putting / intimidating. Being clear and transparent about software licenses is important, however, maybe we should try a friendlier heading? 

Also, the ml5 project has its own license, not MIT! What if we create a new uniform heading something like:

// 👋 Hello! This is an ml5.js made and shared with ❤️.
// Learn more about the ml5.js project: https://ml5js.org/
// ml5.js license and code of conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md

// This example demonstrates _______________
// Any other notes?

shiffman: Tagging in @sproutleaf and @acosmica in case they have any thoughts!

sproutleaf: Agree with your sentiment Dan! I think a new uniform heading with some helpful links would be great.

Should I open an issue to track the progress of heading revamp?

shiffman:
Hello everyone! I'm digging back into ml5.js today and am going to go ahead and merge this! @sproutleaf yes, feel free to start an issue to discuss a "friendly" example heading that we can apply across all examples. @ziyuan-linn I'll open a discussion with the p5.js web editor team re: the API.

PR #65, opened 2023-11-12
add image classifier model

OrpheasK:
Bringing in the ImageClassifier model, this branch picks up from where @gohai made the necessary changes [regarding input normalization](https://github.com/ml5js/ml5-next-gen/commit/98e46956abd891b836556b2d3ad8ef296923ed19). This version alters the way video classification is used, introducing classifyStart() for a similar approach to image input, instead of passing video input as an argument to the constructor. Many thanks to @ziyuan-linn for filling me in on alterations that had taken place while migrating to the next gen version of the library.

PR #71, opened 2024-01-08
Add user options handling

ziyuan-linn:
This PR adds an options handling function, `handleOptions`. Inspired by Mongoose Models, the options handling function takes 2 parameters. The first is a `userObject` from the user, and the second is a `moldObject` that defines how the user input should be filtered. The function returns a `filteredObject` that can be directly passed to the underlying models. 

This function reduces repetitive code and provides an organized way to filter user options objects. To demonstrate this system, the code of handpose has been updated to use the `handleOptions` function. The change in CONTRIBUTING.md has more detailed information about this function.


lindapaiste:
I love this!  Especially the ability to give users very specific warnings or errors.  Excess/ignored properties could be a friendly warning because it's not an error to have extra options, but it might be an indication that they have a typo or used the wrong property name.  Whereas incorrect types would be an error.

src/utils/handleOptions.js
```diff
@@ -0,0 +1,140 @@
...
+  const filteredObject = {};
+
+  for (const key in moldObject) {
+    const userValue = userObject[key];
+    const rules = moldObject[key];
+    const type = evaluate(filteredObject, rules.type);
+    const defaultValue = evaluate(filteredObject, rules.default);
+
+    if (type === "undefined" || type === undefined) {
+      continue;
```
lindapaiste, line 77: Maybe check if the user provided a value and warn them that it is being ignored?

src/utils/handleOptions.js
```diff
@@ -0,0 +1,140 @@
...
+        return enumValue;
+      }
+    }
+  } else {
+    for (const enumValue of enumArray) {
+      if (enumValue === value) {
+        return enumValue;
+      }
+    }
+  }
```
lindapaiste, line 39: I would be inclined to write the loop once and have the `if` condition inside rather than outside.

Alternatively, there are modern JS array methods which can be used to simplify the logic like `.some()`, `.includes()`, and `.find()`.  For example, 
```
// check for exact match
const isValid = enumArray.includes(value);
```
```
// check for case-insensitive match
const matchingVal = enumArray.find(enumValue => enumValue.toLowerCase() === value.toLowerCase());
```

Looks like `userValue` is a typo and should be `value`?

I wonder if enum options in the codebase are always `string` or if there are some `number` enums?

src/utils/handleOptions.js
```diff
@@ -0,0 +1,140 @@
...
+    }
+    // If the user did not provide a value for this option, use the default value.
+    if (userValue === undefined) {
+      filteredObject[key] = defaultValue;
+    }
+    // If the user provided a value for this option, check if it is of the correct type.
+    else if (typeof userValue !== type && type !== "enum") {
+      console.warn(
+        `The value of ${key} is not of type ${type}. Using default value ${defaultValue} instead.`
+      );
```
lindapaiste, line 87: Perhaps only show this warning if the user value is not `undefined` or `null`?

Also you need to be careful about users providing `null` for `object` properties because `typeof null` is `"object"`.

src/Handpose/index.js
```diff
@@ -61,19 +62,42 @@ class Handpose {
...
+        type: (config) =>
+          config.runtime === "mediapipe" ? "string" : "undefined",
+        default: "https://cdn.jsdelivr.net/npm/@mediapipe/hands",
+      },
+      detectorModelUrl: {
+        type: (config) => (config.runtime === "tfjs" ? "string" : "undefined"),
+        default: undefined,
+      },
+      landmarkModelUrl: {
+        type: (config) => (config.runtime === "tfjs" ? "string" : "undefined"),
```
lindapaiste, line 91: If this sort of ternary is showing up everywhere then potentially you could have an additional property to handle that situation.  Like 
```
detectorModelUrl: {
  type: "string",
  default: undefined,
  disallowed: (config) => config.runtime !== "tfjs"
},
```

src/Handpose/index.js
```diff
@@ -61,19 +62,42 @@ class Handpose {
...
-      solutionPath:
-        this.config?.solutionPath ??
-        "https://cdn.jsdelivr.net/npm/@mediapipe/hands",
-      detectorModelUrl: this.config?.detectorModelUrl,
-      landmarkModelUrl: this.config?.landmarkModelUrl,
-    };
-    this.runtimeConfig = {
-      flipHorizontal: this.config?.flipHorizontal ?? false,
-    };
+    const modelConfig = handleOptions(this.config, {
```
lindapaiste, line 65: The options object itself could be defined as a `static` property or outside of the model class.  It doesn't depend of `this` because of how you've designed the conditionals as a function of `config`.  The sort of structured information that you have here is golden for scraping into a documentation site.  We would want to have some sort of system for that where we know that the options config is `Handpose.optionSchema` or whatever.  Not sure what we would do with those functions in the documentation though.

src/utils/handleOptions.js
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * This function takes a userObject and a moldObject as parameters.
+ * Filters the properties of userObject based on the rules defined in moldObject and returns a filteredObject.
+ * Logs out friendly warnings if the properties in userObject did not fulfill the rules.
+ *
+ * @param {object} userObject - options object provided by the user
+ * @param {object} moldObject - an object that defines how the user options object should be filtered
+ * @returns {object} - filtered options object
+ */
+function handleOptions(userObject, moldObject) {
```
lindapaiste, line 67: It might be worth passing the model name (`"handpose"`) as an argument so that it can be printed in error messages.

ziyuan-linn:
@lindapaiste Thank you so much for the review! I will take a look at your suggestions and make some changes.

src/Handpose/index.js
ziyuan-linn: I think this is a good idea!

lindapaiste:
It might be worth reaching out to @Ayush23Dash who worked on the p5.js Friendly Error System (FES) for a GSoC project.  They might have some ideas on how we can integrate with the FES.

https://docs.google.com/document/d/1yL2YAVKJGVIeAmJ_-Np7jF3TYb2IwN2W1Mjk8Epu2Pg/edit
https://github.com/processing/p5.js/tree/main/src/core/friendly_errors

Ayush23Dash:
I think the current state of the Friendly Error Messages is good and provides apt information about what the error is intending to say especially with the implementation of `modelName` as an argument.

src/utils/handleOptions.js
```diff
@@ -0,0 +1,140 @@
+/**
+ * Evaluates the value of a moldObject property.
+ * If the value is a function, it is called with the filtered options object as the its parameter.
+ * Otherwise, the value is returned as is.
+ * @param {object} filteredObject - the current filtered options object
+ * @param {any} value - the value to evaluate
+ * @returns {any} the evaluated value
```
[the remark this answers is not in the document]
lindapaiste: Just for reference.  Not suggesting we do this.  It's probably overkill but I did write an accurate type.  

I tried writing the types using JSDoc but I couldn't get the `@template T` working properly.  So I wrote it in TypeScript.

```
function handleOptions<T extends Record<string, any>>(userObject: T, moldObject: MoldForOptions<T>, modelName: string) {
```

```
interface MoldConfig<D> {
    /**
     * A string defining the correct type.
     */
    type: "number" | "enum" | "boolean" | "string" | "object" | "undefined";
    /**
     * The default value in case the user does not provide a value or provides an erroneous value for the option.
     */
    default: D;
    /**
     * A boolean defining whether the key should be ignored. Defaults to false. Useful when set to a dynamically evaluated value.
     */
    ignore?: boolean;
    /**
     * A number defining the minimum value.
     */
    min?: number;
    /**
     * A number defining the maximum value.
     */
    max?: number;
    /**
     * A boolean defining whether the value should be an integer. Defaults to `false`.
     */
    integer?: boolean;
    /**
     * An array defining a list of valid values.
     */
    enums?: Array<D>;
    /**
     * A boolean defining whether to checks the user value against the enum list case-insensitively. Defaults to `true`.
     */
    caseInsensitive?: boolean;
}

type MoldProperty<PropertyType, OptionsType> = {
    [K in keyof MoldConfig<PropertyType>]?: MoldConfig<PropertyType>[K] | ((options: OptionsType) => MoldConfig<PropertyType>[K])
}

type MoldForOptions<OptionsType> = {
    [K in keyof OptionsType]: MoldProperty<OptionsType[K], OptionsType>;
}
```
[The rest of the types](https://tsplay.dev/w2g49N)

It's a complicated mess 👎 but you do get good Intellisense 👍.

Hovering over a property of the options shows the description of that option, based on what I defined in the `HandposeOptions` interface.
image

Hovering over a field like `default` shows the description of the field and possible values (again, based on `HandposeOptions`).
image

Incorrect configurations give an error.
image

Incorrect options give an error.
image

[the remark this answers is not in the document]
lindapaiste: > I tried writing the types using JSDoc but I couldn't get the `@template T` working properly. So I wrote it in TypeScript.

Turns out it was really simple.  I was trying to put a comment on the same line as the `T` but that was breaking it.

Something like this is more accurate, but also more confusing to anyone who isn't me so idk.
```
/**
 * Evaluates the value of a moldObject property.
 * If the value is a function, it is called with the filtered options object as its parameter.
 * Otherwise, the value is returned as is.
 * @template T
 * @param {object} filteredObject - the current filtered options object
 * @param {T | ((options: object) => T)} value - the value to evaluate
 * @returns {T} the evaluated value
 */
function evaluate(filteredObject, value) {
```

src/utils/handleOptions.js
```diff
@@ -0,0 +1,176 @@
...
+ * @param {any} value - the value to check
+ * @param {Array} enumArray - the array of enum values
+ * @param {boolean} isCaseInsensitive - whether the comparison is case insensitive
+ * @returns {any | undefined} the enum value if the user value is within the enumArray, otherwise undefined
+ */
+function checkEnum(value, enumArray, isCaseInsensitive) {
+  if (isCaseInsensitive && typeof value === "string") {
+    const matchingValue = enumArray.find(
+      (enumValue) => enumValue.toLowerCase() === value.toLowerCase()
+    );
```
lindapaiste, line 59: I'm thinking about numeric enums (ie. `values: [2, 4, 8]`) and if they would break anything.  I think that what you have here is fine because the scenario that throws an error relies on there being nonsense in the `moldObject`, which we control.  A bad user value alone can't break it.  The nonsense way to break it would be using case-insensitive checking with a numeric `enumArray`, and even then it's only an error if the user provides a string: `checkEnum("a", [2, 4, 8], true)`.

shiffman:
This is great work @ziyuan-linn and thank you for the feedback @lindapaiste! @ziyuan-linn I think we can look at merging this next, does anything need to be added now that #65 has been merged?

ziyuan-linn:
@shiffman I just adapted the imageClassifier model to the new options handling system! Since I am not very familiar with the imageClassifier, perhaps @OrpheasK can take a look at my last commit and make sure everything is done correctly. Other than that, I think this is ready to be reviewed and merged.

OrpheasK:
@ziyuan-linn looks good to me! I think it makes sense to merge this version for now and handle further Image Classifier modifications in a new PR using @lindapaiste's discussion thread #83

shiffman:
@ziyuan-linn oops, I think maybe merging #80 caused merge conflicts would you be able to investigate? I probably should have left that for later.

src/utils/handleOptions.js
```diff
@@ -0,0 +1,232 @@
...
+function handleModelName(
+  userValue,
+  possibleValues,
+  defaultValue,
+  ml5ModelName
+) {
+  const modelName = checkEnum(userValue, possibleValues, true);
+  if (userValue === undefined) {
+    return defaultValue;
+  }
```
lindapaiste, line 215: It makes more sense if you swap the order of these two, as there's no need to call `checkEnum` if you're just returning the default.

src/utils/handleOptions.js
```diff
@@ -0,0 +1,232 @@
...
+        filteredObject[key] = userValue;
+      }
+      // Throw an error if the type in moldObject is not "enum", "number", "boolean", "string", or "object"
+      // This should never happen if the moldObject is correctly defined.
+      else {
+        throw new Error(`Unknown type "${type}" for ${key} in moldObject.`);
+      }
+    }
+  }
+  console.log(filteredObject);
```
lindapaiste, line 191: Reminder to remove this before merging.

src/utils/handleOptions.js
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+ * Checks if a value is within the enumArray.
+ * If isCaseInsensitive is true, the comparison is case insensitive.
+ *
+ * @param {any} value - the value to check
+ * @param {Array} enumArray - the array of enum values
+ * @param {boolean} isCaseInsensitive - whether the comparison is case insensitive
+ * @returns {any | undefined} the enum value if the user value is within the enumArray, otherwise undefined
+ */
```
lindapaiste, line 60:
```suggestion
 * @param {string | number} value - the value to check
 * @param {Array<string | number>} enumArray - the array of enum values
 * @param {boolean} isCaseInsensitive - whether the comparison is case insensitive
 * @returns {string | number | undefined} the enum value if the user value is within the enumArray, otherwise undefined
```

src/ImageClassifier/index.js
```diff
@@ -44,36 +54,65 @@ class ImageClassifier {
...
-            break;
-          default:
-            this.modelToUse = null;
-        }
-      } else {
-        // its a url, we expect to find model.json
-        this.modelUrl = modelNameOrUrl;
-        // The teachablemachine urls end with a slash, so add model.json to complete the full path
-        if (this.modelUrl.endsWith("/")) this.modelUrl += "model.json";
+    if (typeof modelNameOrUrl === "string" && isHttpUrl(modelNameOrUrl)) {
```
lindapaiste, line 57: What if it's a relative path?  ie `ml5.imageClassifier("/model/model.json")`

ziyuan-linn: Yeah, that would not work. Thank you for the catch! Though I can't think of a way to cleanly handle this using the `handleModelName` function.

shiffman:
We are meeting and decided now to merge this and leave refinements such as resolving the relative path questions around `modelNameOrUrl` in a future branch. We can track in #83

PR #75, opened 2024-02-21
Add all contributors

ziyuan-linn:
This PR brings in the contributor's list from the old repo.


shiffman:
This is great thank you @ziyuan-linn! If we have people still to add we can do so separately. Also, I can add the bot!

PR #79, opened 2024-02-29
Make use of the built-in tensorflow `topk` function.

lindapaiste:
Resubmission of https://github.com/ml5js/ml5-library/pull/1364

> Kills the `getTopKClassesFromArray` and `getTopKClassesFromTensor` functions and uses the built-in [tfjs `topK` function](https://github.com/tensorflow/tfjs/blob/9769ea46f912a767bd78cce0512782bba718b425/tfjs-core/src/ops/topk.ts) instead. I haven't looked deep into what the `ENGINE.runKernel` is doing under the hood but I would assume it's more performant than our version, which sorts the entire array even though only k results are needed.
> 
> The new `getTopKClasses` function is a wrapper around `tf.topk` which:
> 
> * Accepts both tensors and arrays.
> * Handles the case where no classes are provided by returning the index as the label, eg. `"3"`.
> * Automatically disposes of the input tensor. (This can be disabled via a fourth argument)
> 
> It is used in `darknet`, `doodlenet`, and `speechcommands`.

The original PR also involved a change to the `SoundClassifier`, so this will create a conflict with #28


shiffman:
Thank you @lindapaiste, we discussed this pull request in our meeting this morning and hope to take a look soon. Tagging in @OrpheasK and @gohai for feedback and review!

shiffman:
@OrpheasK is going to look at the merge conflicts here, we would like to focus on getting this integrated into the image classifier (putting sound classifier on pause for the moment) and then once this is merged we can follow the way this is written in the sound classifier.

OrpheasK:
Thanks for merging main @lindapaiste, that has solved the conflicts and makes this ready to merge @shiffman, I will modify the soundClassifier to follow the new topK function according to this and the original PR which seems to include everything necessary. I also understand that all relevant alterations regarding the Image Classifier are taken care of within the PR so no further action needed there!

shiffman:
Wonderful, thank you @lindapaiste! Merging!

PR #80, opened 2024-02-29
Run prettier formatter

ziyuan-linn:
This PR cleans up the formatting of the repo.

- config all text files to use LF line ending
- run prettier formatter on all files


src/utils/generatedImageResult.js
```diff
@@ -16,7 +16,7 @@ import p5Utils from './p5Utils';
  *
  * Accepts options as an object with property `returnTensors` so that models can pass this.config.
```
lindapaiste, line 17: Random but it looks like someone changed the signature of this method and did not change the JSDoc accordingly.  It no longer takes a tensor or accepts options.

shiffman (approved):
Thank you for this @ziyuan-linn!

PR #82, opened 2024-03-02
Set up testing

lindapaiste:
I forked @ziyuan-linn's previous attempts and added a few things to get this _kinda_ working. 
- `BodyPose` tests pass ✔️ 
- `HandPose` tests fail 🚫 but I'm not going down [this rabbit hole](https://github.com/emscripten-core/emscripten/issues/3379) today.

More work still needs to be done here.  The biggest underlying issue is getting browser code to run in Node.js and vice-versa.  Our library is designed to be executed in the browser but these tests are run via Node.
- The polyfills in `setupTests.js` don't seem to apply globally, and need to be repeated in each test file.
    - The `ImageData` one was an issue in the old repo too which is why there is a `polyfillImageData` function in the `testingUtils` file.
  - I don't remember `fetch` being a pain.  I'm looking to see where we used a polyfill and it was only in [one specific test](https://github.com/lindapaiste/ml5-library/blob/7271b3ced723cf142f38d4301896fd759d7addf2/src/FaceApi/index_test.js#L6).
  - TFJS has its own [built-in support](https://github.com/tensorflow/tfjs/blob/f0f981fe306bf548e300536aca485c0ffdd6619e/tfjs-core/src/platforms/platform_node.ts#L20-L61) for `fetch` when using the "node" platform instead of the "browser" platform, but I couldn't figure out how to get that to work.  I could only get it to stick on node if I used `jest-environment-node`, but that created other problems.
  - There was [one test](
https://github.com/lindapaiste/ml5-library/blob/33a6b2d8e3a388f6c591ea63ca409e3e35403351/src/DBSCAN/index.test.js#L21) in the old repo where I explicitly set the TFJS environment to fix an error and that did work.
- We get a bajillion console warnings from TensorFlow, including warnings telling us to use the `@tensorflow/tfjs-node` package which I tried to set up but I still get the warnings.

See also 
https://github.com/ml5js/ml5-library/pull/1345
https://github.com/ml5js/ml5-library/pull/1306


ziyuan-linn:
Thank you @lindapaiste for your help! I was trying to set up testing a while ago but couldn't get it to work.

shiffman:
Yay! This is great, we haven't had any bandwidth to implement testing yet so this is an area that we could use a lot of help with! I don't have a lot of experience with writing tests so don't feel very qualified to review. I'm not sure if @gohai feels the same? Would it make sense to review and merge this with the `BodyPose` tests only or @lindapaiste should we wait until you have the HandPose tests working as well?

lindapaiste:
I think that we should merge it (after resolving conflicts) so that we can continue adding more tests and hope that they will run. 
 TBH I'm pretty good at writing the actual tests but this setup stuff is 🤯.  We can make further changes later.

src/BodyPose/index.test.js
```diff
@@ -1,60 +1,63 @@
...
+      (keypoint) => keypoint.name === "nose"
     );
+    // Should be {"name": "nose", "score": 0.7217329144477844, "x": 454.1112813949585, "y": 256.606980448618}
     expect(nose).toBeTruthy();
-    expect(nose.position.x).toBeCloseTo(448.6, 0);
-    expect(nose.position.y).toBeCloseTo(255.9, 0);
-    expect(nose.score).toBeCloseTo(0.999);
+    expect(nose.x).toBeCloseTo(454.1, 0);
+    expect(nose.y).toBeCloseTo(256.6, 0);
+    expect(nose.score).toBeCloseTo(0.721, 2);
```
lindapaiste, line 52: As an aside - It is slightly concerning that the old model had a `0.999` confidence score for detecting the nose and the new model is only `0.721`.

shiffman:
> Is support for running ml5.js in Node.js something that's on the agenda? If so then we should stick with Jest because a lot of the issues that come up in these tests are genuine issues that would come up when using the library.

@lindapaiste Speaking for myself, this is something that I would like to see in the future, but it is not something I would prioritize in the near term. For me, having 1.0 out and released with updated models and refreshed documentation for use with p5.js client-side is the top priority. I think we could revisit the node.js question after that has happened. 

Let me know when you want me to review for merging! Maybe it makes sense to merge with 1 or 2 tests only and then others can contribute additional tests or updates in separate branches. I don't have an opinion on jest or karma as I don't really have experience with either! If you are excited to start with a platform I would say go for it!

lindapaiste:
@shiffman I think let's go ahead and merge this as-is right now.  We can continue to make improvements in other branches/PRs.

src/BodyPose/index.test.js
gohai: I am wondering if the confidence score will change when running it with the _tfjs_ runtime rather than _mediapipe_ 🤔

shiffman: Yes, maybe it would make more sense to just validate that a floating point confidence score is outputted rather than a specific value? But we can address this later!

shiffman:
Excited to merge this! I'll start playing with the tests more and we can incrementally improve them and add more. I can look into how to get them running as CI checks on github too!

gohai:
@lindapaiste Are those `.babelrc` and `babel.config.json` files in the top-level directory strictly necessary for the test infrastructure to work? (Always a bit deterring encountering large number of dot or configuration files as the first thing one looks at.)

Also wondering if `setupTests.js` might also want to go a folder `tests` or `testing` (or whatever is most common for those)?

lindapaiste:
> @lindapaiste Are those `.babelrc` and `babel.config.json` files in the top-level directory strictly necessary for the test infrastructure to work? (Always a bit deterring encountering large number of dot or configuration files as the first thing one looks at.)

We can [move it into package.json](https://babeljs.io/docs/configuration#packagejson).

> Also wondering if `setupTests.js` might also want to go a folder `tests` or `testing` (or whatever is most common for those)?

Sure.  It might also be cleaner/clearer if we move the test files into that folder, instead of putting them alongside the source code.

gohai:
Thank you @lindapaiste

Issue #78, opened 2024-02-23
Topology mismatch with 0.20.0.alpha releases


j3nsykes:
Not wanting to open duplicate issues but I have narrowed [this issue](https://github.com/ml5js/ml5-library/issues/1479)
to be related to the last 3 alpha releases of ML5. 

It is not present with ml5@latest/dist/ml5.min.js

However, it seems when saving and loading the files with the neural net function in the alpha releases the tensorflow topology is expecting a different structure ?

I have tried this with nextgen handpose and also a simple mouse X,Y regression. Both cause this error as soon as the version is 0.20.0.alpha 1,2 or 3.

shiffman:
Thank you for reporting this @j3nsykes! Can you share one of your examples where you are getting this error. I'd like to take a look. I have been using the `ml5.neuralNetwork()` class for my new chapters in The Nature of Code, but I never actually tested the saving and loading model features! 

it's absolutely fine to open duplicate issues if they are present in this new repo! I realize we've caused some extra confusion and additional labor in moving to a new repo, but it helped us re-engage with the project after a dormant period! Hopefully we'll be able to bring everything together smoothly into a new release soon!!!

In case you are curious to take a look: https://natureofcode.com/neural-networks/ https://natureofcode.com/neuroevolution/)

j3nsykes:
Hey Dan, 
Sure its happening here when I'm adapting one of [Golan's hand tracking regression examples](https://editor.p5js.org/jen_GSA/sketches/DYTA7MGjB) to be saved and loaded but also in a [simple mouse regression](https://editor.p5js.org/jen_GSA/sketches/r5Ncmj7Jv) example too (so not related to handpose.js)

The mouse example I've been using to test which version of ML5 it breaks with.

shiffman:
Ok, so I've dug into this error a bit more! The code still works if you are loading the model files directly with a single filename rather than an object:

This works:
```js
  classifier.load("model.json", modelLoaded);
```

However, this does not work:
```js
  const modelDetails = {
    model: "model.json",
    metadata: "model_meta.json",
    weights: "model.weights.bin",
  };
  classifier.load(modelDetails, modelLoaded);
```

However, only the second option is compatible with the web editor where the files are stored separately and the code needs to extrapolate the hosted path (?). It does that via something along the lines of:

```js
 this.model = await tf.loadLayersModel(tf.io.browserFiles([file1, file21]));
 ```
 
 Rather than the more simple:
 
 ```js
[...]

PR #84, opened 2024-03-04
Fix nerural network load model topology mismatch

ziyuan-linn:
This PR removes the `JSON.stringify` call on `modelJsonResult.data` and `modelMetadata.data` since Axios already returns those as strings.

This addresses issue #78.


src/NeuralNetwork/NeuralNetwork.js
```diff
@@ -251,7 +251,7 @@ class NeuralNetwork {
       const modelJsonResult = await axios.get(filesOrPath.model, {
         responseType: "text",
```
lindapaiste, line 252: The response `.data` is already a string specifically because of the `responseType: "text"` here.  If you don't specify a `responseType` then `axios` will default to parsing it as JSON.

src/NeuralNetwork/NeuralNetworkData.js
```diff
@@ -839,8 +839,7 @@ class NeuralNetworkData {
       let modelMetadata = await axios.get(filesOrPath.metadata, {
         responseType: "text",
       });
-      modelMetadata = JSON.stringify(modelMetadata.data);
-      modelMetadata = JSON.parse(modelMetadata);
+      modelMetadata = JSON.parse(modelMetadata.data);
```
lindapaiste, line 842: If we want it as JSON then I think you could just remove the `responseType: "text"` and let `axios` parse it.  Try this:
```
const metadataRes = await axios.get(filesOrPath.metadata);
this.meta = metadataRes.data;
```

src/NeuralNetwork/NeuralNetwork.js
```diff
@@ -251,7 +251,7 @@ class NeuralNetwork {
...
       });
-      const modelJson = JSON.stringify(modelJsonResult.data);
+      const modelJson = modelJsonResult.data;
       // TODO: browser File() API won't be available in node env
       const modelJsonFile = new File([modelJson], "model.json", {
```
lindapaiste, line 256: TBH this entire section of the code feels a bit odd and unnecessary.  We are fetching the files, creating `File` objects, and then passing them to the TF model with `tf.io.browserFiles`.  But TF is definitely capable of fetching files on its own so it seems like we shouldn't have to do that on our end.  It seems like we should just be able to pass the URLs.  That's a discussion for another time though, because maybe there's a reason that this is necessary and I'm just not seeing it.

shiffman: I believe this is necessary due to the way that the p5.js web editor stores the model files.

src/NeuralNetwork/NeuralNetworkData.js
shiffman: Thank you @lindapaiste! @ziyuan-linn take a look at this feedback and after you make any changes as you see fit I can merge!

src/NeuralNetwork/NeuralNetwork.js
lindapaiste: Yeah that makes sense.  I know the p5.js web editor well and the way that it handles uploaded files is complicated and messy because there is a URL that the user sees and uses in their code, `"/model/model.json"`, and then there is the actual URL of the file which is some long string file name on an S3 server.  So when TF is looking for other files with specific names in the same directory it's not gonna find them.  We might be able to define a custom TF loader, but it's not as simple as I was thinking.

shiffman:
Merging! If there are further refinements we can make to simplify the file loading but retain compatibility with the web editor, we can do so in a new branches!

PR #89, opened 2024-03-14
🧹 NeuralNetwork cleanup pt. 1 -- callbacks

lindapaiste:
Reboot of https://github.com/ml5js/ml5-library/pull/1407

---

Fixes https://github.com/ml5js/ml5-library/issues/1343
* Change typo `this.createMetadata` to `this.createMetaData`.

---

Cleanup
* Remove all callback arguments from the methods of internal classes `NeuralNetwork` and `NeuralNetworkData` and use async methods instead.  The callbacks can be handled by the outer `DiyNeuralNetwork` class when it calls these methods.
* Call async functions pairs (`save`/`saveMeta`, `load`/`loadMeta`) in parallel instead of sequentially.
* Return a promise when no callback is provided in:
  * nn.save()
  * nn.load()
  * nn.loadData()
  * nn.train()
* Add callback and promise handling to:
  * nn.saveData()
* Fix inconsistent handling of `this.ready` where it was sometimes a `Promise` and sometimes a `boolean` `true`.  Now it is always a `Promise` which resolves to the instance.
* ~~Fix `else if` in `init()` which prevented loading both a `dataUrl` and a `modelUrl`.~~ (forgot to do this, but I put it in another PR)
* Delete unecessary `loadDataInternal()` and call `loadDataFromUrl()` directly.
* Delete unecessary instance property `this.callback`. (Note: a few other models do this too)
* Add lots of JSDoc type and comments.
* Change the JSDoc types on the `callCallback` function to match the new value-first style.


shiffman:
@lindapaiste this is fantastic work, thank you! I have reviewed the code changes and they all look good to me. I've also manually tested all of my examples and they run as expected. The only thing I was not able to test was that the saving and loading of the model still works, do you have an example that you tested that with?

@ziyuan-linn I know your work was largely just bringing these class over, but would you mind taking a review of this branch as well before I merge to see if anything doesn't match with how you are maintaining other related models?

ziyuan-linn (approved):
@lindapaiste thank you so much! I like these changes and don't see any issues. 

I tested the saving and loading feature and they still work as expected.

ziyuan-linn:
I can also work on some training, saving, and loading examples.

lindapaiste:
@ziyuan-linn Thank you for testing it! To be honest I did not test out the saving and loading.

lindapaiste:
This also fixes a bug which I just discovered where if you don't have a `dataUrl` or `modelUrl` then the `callback` will NEVER called.

shiffman:
Hi @lindapaiste looks like some conflicts arose related to the `preload()` cleanup!

shiffman:
Great! I re-tested all the examples and everything is working great.

PR #91, opened 2024-03-16
🧹 NeuralNetwork cleanup - passing arguments between methods

lindapaiste:
A. There are a lot of method calls in the NeuralNetwork classes which pass around variables that are properties of `this`.  If the argument is the same every time that the method is called then it doesn't need to be an argument -- instead the method can access the property itself.

Removed `dataRaw` argument from:
- prepareForTraining
- normalizeData
- normalizeDataRaw
- normalizeInputData
- getDataOneHot
- getInputMetaOneHot
- applyOneHotEncodingsToDataRaw
- convertTrainingDataToTensors
- createMetadata & createMetaData
- getDTypesFromData
- getDataStats
- getInputMetaStats
- getDataUnits
- getInputMetaUnits

Removed other arguments from:
- formatInputsForPredictionAll (`meta` and `inputHeaders`)
- formatInputsForPrediction (`meta` and `inputHeaders`)
- createNetworkLayers (`meta`)
- addDefaultLayers (`task` and `meta`)
- compile (`_modelOptions` and `_learningRate`)

---

B. Removed `return` statements that were not actually needed from a bunch of methods.  These values were either not used already, or were only used for arguments which I removed. Generally speaking, a method should either modify `this` or return a value but not both.

Removed the return value from:
- loadCSV
- loadJSON
- loadBlob
- loadDataFromUrl
- formatRawData
- prepareForTraining
- normalizeData
- createMetadata
- getDTypesFromData
- getDataStats
- getDataOneHot
- getDataUnits

---

C. Removed the whole block of `.bind(this);` from the `NeuralNetworkData` constructor.  The only benefit of this sort of binding is that it allows you to pass the method around as a variable without losing the `this` context.  For example, you can write `input.map(this.formatInputItem)` instead of `input.map((item) => this.formatInputItem(item))`.  But we don't do that anywhere in the code.  We can leave the `.bind` in on the public methods of the `DiyNeuralNetwork` class to support that sort of usage by end-users.  For all private and internal methods it's useless.

---

D. Removed some unnecessary cloning/copying of variables, such as `const meta = Object.assign({}, this.meta);`.  This creates a shallow copy.  So when we then use `meta.inputs` it is a reference to the same object as `this.meta.inputs`.  That's not a problem but it means that the cloning doesn't accomplish anything and should be removed.

### DIFF
--- a/src/NeuralNetwork/NeuralNetwork.js
+++ b/src/NeuralNetwork/NeuralNetwork.js
@@ -57,11 +57,11 @@ class NeuralNetwork {
   /**
    * add layer to the model
    * if the model has 2 or more layers switch the isLayered flag
-   * @param {*} _layerOptions
+   * @param {tf.Layer} layer
+   * @void
    */
-  addLayer(_layerOptions) {
-    const LAYER_OPTIONS = _layerOptions || {};
-    this.model.add(LAYER_OPTIONS);
+  addLayer(layer) {
+    this.model.add(layer);
 
     // check if it has at least an input and output layer
     if (this.model.layers.length >= 2) {
--- a/src/NeuralNetwork/NeuralNetworkData.js
+++ b/src/NeuralNetwork/NeuralNetworkData.js
@@ -21,43 +21,6 @@ class NeuralNetworkData {
     this.data = {
       raw: [], // array of {xs:{}, ys:{}}
     };
-
-    // methods
-    // summarize data
-    this.createMetadata = this.createMetadata.bind(this);
-    this.getDataStats = this.getDataStats.bind(this);
-    this.getInputMetaStats = this.getInputMetaStats.bind(this);
-    this.getDataUnits = this.getDataUnits.bind(this);
-    this.getInputMetaUnits = this.getInputMetaUnits.bind(this);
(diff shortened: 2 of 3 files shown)


shiffman:
I love this cleanup so much! I often have found it quite confusing when trying to debug the neural network classes with a lot of the redundancies and multiple instances of the same variables. @ziyuan-linn would you have time to look this over as well as #89?

ziyuan-linn:
I was wondering why the blocks of `.bind(this)` was there😂, @lindapaiste your explanation was great. I honestly don't even think this would be used much by the end-users, but I guess there is no harm leaving it.

ziyuan-linn (approved):
@lindapaiste Thank you! These changes looks great. I tested all existing examples as well as saving and loading and everything works as expected.

lindapaiste:
> I was wondering why the blocks of `.bind(this)` was there😂, @lindapaiste your explanation was great. I honestly don't even think this would be used much by the end-users, but I guess there is no harm leaving it.

I would be on board with deleting all of them TBH.

shiffman:
Thank you @lindapaiste!

PR #92, opened 2024-03-16
Rename last remnants of BodyPix

gohai:
This fixes the body segmentation examples in #81.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,7 +7,7 @@ import * as tf from "@tensorflow/tfjs";
 import * as tfvis from "@tensorflow/tfjs-vis";
 import p5Utils from "./utils/p5Utils";
 import setBackend from "./utils/setBackend";
-import bodyPix from "./BodySegmentation";
+import bodySegmentation from "./BodySegmentation";
 import communityStatement from "./utils/communityStatement";
 import imageClassifier from "./ImageClassifier";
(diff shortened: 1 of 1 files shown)


ziyuan-linn:
@gohai Thank you for fixing this issue! Looks like somehow `bodySegmentation` got reverted backed to `bodyPix` during one of the merges.

documentation.md
```diff
@@ -22,7 +22,7 @@ const bodySegmentation = ml5.bodySegmentation(?modelName, ?options, ?callback);
 
 **Parameters:**
 
-- **modelName**: OPTIONAL: A string specifying which model to use, "SelfieSegmentation" or "BodyPix".
```
ziyuan-linn, line 25: I don't think this needs to be changed since the `modelName` parameter can still be set to either `SelfieSegmentation` or `BodyPix`. 

Here is the current `handleModelName` function for `bodySegmentation`:
```js
this.modelName = handleModelName(
      this.modelName,
      ["BodyPix", "SelfieSegmentation"], //<--- possible values for modelName
      "SelfieSegmentation", //<-- default value of modelName
      "bodySegmentation" //<-- name of the ml5 model, used by the friendly error system
);
```

gohai: Thanks @ziyuan-linn - fixed.

shiffman:
Thank you @gohai and @ziyuan-linn!

PR #93, opened 2024-03-16
Create tests for modelLoader.js + fix relative path handling

lindapaiste:
This PR adds tests for the URL-handling utilities in the `modelLoader.js` file.

I also fixed some issues that I found due to failures in tests.  When I wrote the `ModelLoader` class [I didn't modify the `getModelPath` function](https://github.com/ml5js/ml5-library/commit/674c4759c86580ad159006f0dca10398d70e64a4#diff-9ab9ff082c354557f42d0cbc28bae4afacb65dff78e1b150e24206d8a0deb101) (which I have now renamed to `getAbsolutePath` because that is a more descriptive name of what it does).  This function wasn't really doing the right thing because it was using `window.location.pathname` which does not include the domain.  It also wasn't looking for trailing or leading slashes, so it could wind up with no `/` or with a double `//`.  The revised version can now properly handle paths which are relative to the root (`'/folder/model.json'`) or relative to the current page (`'folder/model.json'`).  It can NOT handle more complex relative paths, like `'../folder/model.json'`.

#### Follow-up
Right now we are by default converting all relative URLs to absolute URLs.  We should see if this is actually necessary and/or beneficial.  If we do nothing, can TFJS load a file from `'../folder/model.json'`?

### DIFF
--- a/src/utils/modelLoader.js
+++ b/src/utils/modelLoader.js
@@ -18,9 +18,16 @@ export function isAbsoluteURL(str) {
  * @param {string} absoluteOrRelativeUrl
  * @returns {string}
  */
-export function getModelPath(absoluteOrRelativeUrl) {
+export function getAbsolutePath(absoluteOrRelativeUrl) {
   if (!isAbsoluteURL(absoluteOrRelativeUrl) && typeof window !== "undefined") {
-    return window.location.pathname + absoluteOrRelativeUrl;
+    // If it starts with "/" then it is relative to the domain/subdomain.
+    if (absoluteOrRelativeUrl.startsWith('/')) {
+      return window.location.origin + absoluteOrRelativeUrl;
+    }
+    // Otherwise it is relative to the current page.
+    let base = window.location.href;
+    if (!base.endsWith('/')) base += '/';
+    return base + absoluteOrRelativeUrl;
   }
   return absoluteOrRelativeUrl;
 }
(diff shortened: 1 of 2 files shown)


lindapaiste:
here is an issue where someone got an error when trying to use a '../folder' path: https://github.com/ml5js/ml5-library/issues/1442

shiffman:
This is a great change, thank you! @ziyuan-linn can you test as a double-check?

ziyuan-linn (approved):
The tests look good to me! We can also look into more sophisticated URL handling in the future, or just let tfjs handle it if possible.

Just wondering, is `modelLoader` currently being used anywhere? Seems like this is not hooked up to `neuralNetwork`?

shiffman:
I'm going to go ahead and merge this, @ziyuan-linn feel free to open up a new issue with anything from https://github.com/ml5js/ml5-next-gen/pull/93#pullrequestreview-1956216389 if you feel it merits further discussion elsewhere!

lindapaiste:
> Just wondering, is `modelLoader` currently being used anywhere? Seems like this is not hooked up to `neuralNetwork`?

Looks like it's only being used in `Sentiment` at the moment 😢.  We can and should use it in the `ImageClassifier`.

I need to make some tweaks in order to support the `neuralNetwork`:
1. To support passing all 3 URLs at once in a dictionary object (we don't do that anywhere else)
2. Because for some reason we use the file name `model_meta.json` instead of `metadata.json`.  I don't know why, maybe we purposefully wanted it to not be confused with a TFJS `metadata.json` file.  But that sort of inconsistency makes it hard to write clean code.

PR #94, opened 2024-03-16
Write tests for `handleOptions.js`

lindapaiste:
This one is mostly self-explanatory if you look at the `it` and `describe` texts.  Checks:
- general returning of default values
- `ignore` handling
- enum handling
- number handling

Added `resetMocks: true` to the Jest config to automatically reset mocked functions after every test.


shiffman:
yay, excited to try these tests!

PR #95, opened 2024-03-17
🧹 Cleanup Neural Network task-dependent logic

lindapaiste:
Reboot of the changes in this commit: https://github.com/ml5js/ml5-library/pull/1410/commits/f8e4b8f95079e498c2b6c010f793dbc52db4c3b0
from https://github.com/ml5js/ml5-library/pull/1410

I branched this on top of the previous 2 PRs, expecting them to be merged first.

The idea is essentially that all logic which depends on which task is selected (`classification`, `regression` or `imageClassification`) is all declared in one place in a new `getTask` file.  This makes it a lot easier to see how we are handling each situation.  And easier to potentially add more task presets in the future.

The rest of the notes copied from the previous PR:

---

Introduces a level of abstraction for the "task" of the neural network:

`getTask.ts`
- Define an interface `NNTask` which specifies what information a task needs to provide:
  - Its default layers.
  - Its compile options.
  - Any overrides to the NN default options.
  - How to create sample data to use when option `neuroEvolution` is set.
  - Its `name` (not currently not used, possible 🪓)
- Create three concrete implementations of that interface for the three tasks: `classification`, `regression`, and `imageClassification`.
  - `imageClassification` uses some of the same functions as `classification`.
  - `getSampleData` is only possible in a limited set of circumstances (see issue #1409), so I'm throwing a lot of specific errors when those circumstances aren't met.
- Export a function `getTask` to convert a `string` task name into a `NNTask` object.

`NeuralNetwork/index.ts`
- Constructor sets an instance property `this.task` with the task object based on `options.task`.  This defaults to `regression` which seems to match what was in the default layers before.
- Remove large blocks of code in `addDefaultLayers()` and `compile()` which was moved into the task objects.  Now we call a function on `this.task`:
  - `const layers = this.task.createLayers(inputUnits, hiddenUnits, outputUnits);`
  - `const options = this.task.getCompileOptions(this.options.learningRate);`

`NeuralNetwork/NeuralNetwork.ts`
- I'm like 99% sure that using `optimizer.call(this, learningRate)` to set the `thisArg` on the optimizer doesn't actually make a difference? (Please correct me if I am wrong).  I removed `setOptimizerFunction()`, which, despite the name, just creates an optimizer and doesn't set anything.  The instantiation of the optimizer is handled in the `getCompileOptions()` function of the task, which gets the `learningRate` as an argument.

### DIFF
--- a/src/NeuralNetwork/index.js
+++ b/src/NeuralNetwork/index.js
@@ -24,8 +24,7 @@ const DEFAULTS = {
   neuroEvolution: false,
 };
 class DiyNeuralNetwork {
-  constructor(options, cb) {
-    this.callback = cb;
+  constructor(options, callback) {
 
     // Is there a better way to handle a different
     // default learning rate for image classification tasks?
@@ -47,14 +46,11 @@ class DiyNeuralNetwork {
       training: [],
     };
 
-    this.ready = false;
-
     // Methods
     this.init = this.init.bind(this);
     // adding data
     this.addData = this.addData.bind(this);
     this.loadDataFromUrl = this.loadDataFromUrl.bind(this);
-    this.loadDataInternal = this.loadDataInternal.bind(this);
     // metadata prep
     this.createMetaData = this.createMetaData.bind(this);
     // data prep and handling
@@ -99,7 +95,7 @@ class DiyNeuralNetwork {
     this.crossover = this.crossover.bind(this);
 
     // Initialize
-    this.init(this.callback);
+    this.ready = callCallback(this.init(), callback);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


src/NeuralNetwork/index.js
```diff
@@ -591,20 +538,13 @@ class DiyNeuralNetwork {
...
-    if (!this.neuralNetwork.isLayered) {
-      this.options.layers = this.addDefaultLayers(
-        this.options.task,
-        this.neuralNetworkData.meta
-      );
+      if (this.options.layers && this.options.layers.length > 2) {
+        this.createNetworkLayers(this.options.layers);
+      } else {
+        this.addDefaultLayers();
+      }
```
lindapaiste, line 547: I removed the modification of `this.options.layers` because it felt weird/sloppy/wrong.  Now I'm slightly concerned that this change could create problems with the `.copy()` method, which uses `this.options` to create the cloned neural network.  Let's put this PR on hold for right now until I have a change to look into it more.

shiffman: Thanks for keeping a close eye on this @lindapaiste! The `copy()` method is critical for the neuroevolution examples I developed for the nature of code book.

Issue #90, opened 2024-03-16
Error in neuralNetrwork "mouse gesture" example if you click without dragging


lindapaiste:
It works fine if you drag as intended, but there's no instructions telling the user what to do.

If you click on the canvas (without ever having dragged) then you will get a crash triggered by this line:
https://github.com/ml5js/ml5-next-gen/blob/13ba73c556a9b44baf0025be5599073e3697d6c9/examples/NeuralNetwork-mouse-gesture/sketch.js#L79

The `end` variable is set by the `mouseDragged` function.  So if there is no drag then `end` is `undefined` and the `p5.Vector.sub()` function will crash.

> Uncaught TypeError: Cannot read properties of undefined (reading 'copy')
>    at Function.sub (p5.js:92156:31)
>    at mouseReleased (sketch.js:79:23)
>    at _main.default._onmouseup (p5.js:78236:38)

https://github.com/ml5js/ml5-next-gen/blob/13ba73c556a9b44baf0025be5599073e3697d6c9/examples/NeuralNetwork-mouse-gesture/sketch.js#L74-L84

PR #99, opened 2024-03-20
fix mouse gesture example click without drag bug

ziyuan-linn:
This PR fixes the bug of #90 by recoding the end position in addition to the start position when `mousePressed` is called.


shiffman:
Ah, thanks, this is a good fix!

PR #100, opened 2024-03-20
Resize body pose bounding box

ziyuan-linn:
The PR adds the `resizeBoundingBoxes` function, which resizes the bounding box values of `MoveNet` models relative to the media size.

I am also fixing some typos in the `addKeypoints` method where `hand` is supposed to be `pose`.


shiffman:
This is great, thank you @ziyuan-linn!

PR #101, opened 2024-03-20
Remove .prettierrc.json (from root directory)

gohai:
"yarn format" works the same without, so unless we have a need to overwrite its default settings, I'd like to propose getting rid of it. (People going to the repo page will have to scroll through everything in the root directory before they get to the Readme.)

(sorry for going full OCD with this one)


ziyuan-linn:
My understanding is that without an empty `.prettierrc.json` file, Prettier will use the userspace configurations to run `yarn format` which might be different for everyone (I could be very wrong about this).

Perhaps we can move it somewhere else? I think Pettier also accepts a `prettier` key in `package.json`.

lindapaiste:
> My understanding is that without an empty `.prettierrc.json` file, Prettier will use the userspace configurations to run `yarn format` which might be different for everyone (I could be very wrong about this).

When I run `format` I get 65 changed files.  The diff shows "Contents have differences only in line separators" on most.  So I think that we do want to set an `"endOfLine"` option.

gohai:
Looks like prettier doesn't take a global configuration file into consideration - so I believe this could be deleted:

> Prettier intentionally doesn’t support any kind of global configuration. This is to make sure that when a project is copied to another computer, Prettier’s behavior stays the same.

But let me add a stub config to our `package.json` so that we can add the options as @lindapaiste is suggesting there.

PR #104, opened 2024-03-21
bodyPose skeleton examples

ziyuan-linn:
This PR adds example p5 sketches that draw skeletons for the `MoveNet` and `BlazePose` models. 

~~The [nested arrays that define the skeletal connections](https://github.com/tensorflow/tfjs-models/blob/c5fcabec4cc0335bdfabeb14dc2adfd6ef0cde8c/pose-detection/src/constants.ts#L72) are copied into the p5 sketches and iterated over to draw the skeleton lines.~~

As @lindapaiste suggested, I added a new `getSkeleton` function to `bodyPose` that re-exports the skeleton data from tfjs. There's probably a better name than `getSkeleton` though.

### DIFF
--- a/examples/BodyPose-blazepose-skeleton/index.html
+++ b/examples/BodyPose-blazepose-skeleton/index.html
@@ -9,7 +9,7 @@
   <head>
     <meta charset="UTF-8" />
     <title>ml5.js BodyPose BlazePose Skeleton Example</title>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.9.0/p5.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.9.2/p5.js"></script>
     <script src="../../dist/ml5.js"></script>
   </head>
 
--- a/examples/BodyPose-blazepose-skeleton/index.html
+++ b/examples/BodyPose-blazepose-skeleton/index.html
@@ -1,0 +1,19 @@
+<!--
+ Copyright (c) 2024 ml5
+ 
+ This software is released under the MIT License.
+ https://opensource.org/licenses/MIT
+-->
+
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <title>ml5.js BodyPose BlazePose Skeleton Example</title>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.9.0/p5.js"></script>
+    <script src="../../dist/ml5.js"></script>
+  </head>
+
+  <body>
+    <script src="sketch.js"></script>
+  </body>
+</html>
--- a/examples/BodyPose-blazepose-skeleton/sketch.js
+++ b/examples/BodyPose-blazepose-skeleton/sketch.js
@@ -6,45 +6,7 @@
 let video;
 let bodyPose;
 let poses = [];
-
-let connections = [
-  [0, 1],
-  [0, 4],
-  [1, 2],
-  [2, 3],
-  [3, 7],
-  [4, 5],
-  [5, 6],
-  [6, 8],
-  [9, 10],
-  [11, 12],
-  [11, 13],
-  [11, 23],
-  [12, 14],
-  [14, 16],
-  [12, 24],
-  [13, 15],
-  [15, 17],
-  [16, 18],
-  [16, 20],
-  [15, 17],
-  [15, 19],
-  [15, 21],
-  [16, 22],
-  [17, 19],
-  [18, 20],
-  [23, 25],
-  [23, 24],
-  [24, 26],
-  [25, 27],
-  [26, 28],
-  [27, 29],
-  [28, 30],
-  [27, 31],
-  [28, 32],
-  [29, 31],
-  [30, 32],
-];
+let connections;
 
 function preload() {
   // Load the bodyPose model
@@ -61,6 +23,8 @@ function setup() {
 
   // Start detecting poses in the webcam video
   bodyPose.detectStart(video, gotPoses);
+  //get the skeleton connection information
+  connections = bodyPose.getSkeleton();
 }
 
 function draw() {
--- a/examples/BodyPose-blazepose-skeleton/sketch.js
+++ b/examples/BodyPose-blazepose-skeleton/sketch.js
@@ -1,0 +1,101 @@
+// Copyright (c) 2024 ml5
+//
+// This software is released under the MIT License.
+// https://opensource.org/licenses/MIT
+
+let video;
+let bodyPose;
+let poses = [];
+
+let connections = [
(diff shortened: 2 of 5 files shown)


lindapaiste:
What happens if the image does not include the entire body?  Could we end up with cases where `pointA` or `pointB` are `undefined`? (leading to `TypeError: cannot read property 'x' of undefined`)

> The [nested arrays that define the skeletal connections](https://github.com/tensorflow/tfjs-models/blob/c5fcabec4cc0335bdfabeb14dc2adfd6ef0cde8c/pose-detection/src/constants.ts#L72) are copied into the p5 sketches and iterated over to draw the skeleton lines.

I wonder if it makes sense for us to re-export the [utils](https://github.com/tensorflow/tfjs-models/blob/c5fcabec4cc0335bdfabeb14dc2adfd6ef0cde8c/pose-detection/src/util.ts#L32) (`getAdjacentPairs`) as properties of `ml5.bodyPose` or something like that.

ziyuan-linn:
> What happens if the image does not include the entire body? Could we end up with cases where `pointA` or `pointB` are `undefined`? (leading to `TypeError: cannot read property 'x' of undefined`)

From my understanding, `MoveNet` will always output 17 keypoints, and `BlazePose` will always output 32 keypoints when a person is detected. The confidence score for some of the keypoints will be very low if the image is a partial body.

For `BlazePose`, those low-confidence keypoints will usually be off-canvas so nothing really needs to be done. For `MoveNet`, the keypoints tend to bunch up at the edge of the canvas where the body is cropped off, causing a mess. In that case, the example will not draw the skeleton line if `pointA` or `pointB` has low confidence.

ziyuan-linn:
> I wonder if it makes sense for us to re-export the [utils](https://github.com/tensorflow/tfjs-models/blob/c5fcabec4cc0335bdfabeb14dc2adfd6ef0cde8c/pose-detection/src/util.ts#L32) (`getAdjacentPairs`) as properties of `ml5.bodyPose` or something like that.

I think this is a great idea! I can look into implementing this.

examples/BodyPose-blazepose-skeleton/index.html
```diff
@@ -0,0 +1,19 @@
+<!--
+ Copyright (c) 2024 ml5
```
shiffman, line 2: just commenting about the headers again!

examples/BodyPose-blazepose-skeleton/index.html
```diff
@@ -0,0 +1,19 @@
...
+ 
+ This software is released under the MIT License.
+ https://opensource.org/licenses/MIT
+-->
+
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <title>ml5.js BodyPose BlazePose Skeleton Example</title>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.9.0/p5.js"></script>
```
shiffman, line 12:
```suggestion
    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.9.2/p5.js"></script>
```

let's try to keep everything updated to latest p5.js (esp due to the mirroring feature)

examples/BodyPose-skeleton/index.html
```diff
@@ -0,0 +1,19 @@
...
+ 
+ This software is released under the MIT License.
+ https://opensource.org/licenses/MIT
+-->
+
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <title>ml5.js BodyPose Skeleton Example</title>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.9.0/p5.js"></script>
```
shiffman, line 12:
```suggestion
    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.9.2/p5.js"></script>
```

examples/BodyPose-blazepose-skeleton/sketch.js
```diff
@@ -0,0 +1,65 @@
...
+
+function draw() {
+  // Draw the webcam video
+  image(video, 0, 0, width, height);
+
+  //draw the skeleton connections
+  for (let i = 0; i < poses.length; i++) {
+    let pose = poses[i];
+    for (let j = 0; j < connections.length; j++) {
+      let pointAIndex = connections[j][0];
```
shiffman, line 38: I wonder if there is a naming convention that might be more clear than using a 2D array (which can be very confusing for beginners as well as not being descriptive). would this be better:

```js
let pointAIndex = connections[j].a
let pointAIndex = connections[j].b
```

Is there something more descriptive? Also, we'll need to add this to the documentation

examples/BodyPose-skeleton/sketch.js
```diff
@@ -0,0 +1,70 @@
...
+
+function draw() {
+  // Draw the webcam video
+  image(video, 0, 0, width, height);
+
+  //draw the skeleton connections
+  for (let i = 0; i < poses.length; i++) {
+    let pose = poses[i];
+    for (let j = 0; j < connections.length; j++) {
+      let pointAIndex = connections[j][0];
```
shiffman, line 38: same note here at 2D arrays

examples/BodyPose-skeleton/sketch.js
```diff
@@ -0,0 +1,70 @@
...
+      let pointBIndex = connections[j][1];
+      let pointA = pose.keypoints[pointAIndex];
+      let pointB = pose.keypoints[pointBIndex];
+      // Only draw a line if both points are confident enough
+      if (pointA.score > 0.1 && pointB.score > 0.1) {
```
shiffman, line 43: Are we using `score` vs. `confidence` consistently across all models? This is something we should be consistent about. What does the `imageClassifier` use for example?

shiffman (commented):
Left some comments!

examples/BodyPose-blazepose-skeleton/sketch.js
lindapaiste: It might help to add an intermediate variable?
```
let currentConnection = connections[j];
let pointAIndex = currentConnection[0];
let pointBIndex = currentConnection[1];
```

(If this was not for beginners I would definitely use destructuring `let [pointAIndex, pointBIndex] = connections[j]`)

examples/BodyPose-blazepose-skeleton/index.html
ziyuan-linn: Yes, we should also update these on all other examples!

shiffman:
In talking with @ziyuan-linn I realized that the array pairs are coming from tf.js and I don't think we should rewire them to be object properties `a` and `b`, we can always adjust the examples like how @lindapaiste suggested to make them easier to read.

Issue #76, opened 2024-02-21
Header for Examples


shiffman:
This is maybe a silly comment, but I sometimes find that when looking at the examples that the copyright notice at the top is a little off-putting / intimidating. Being clear and transparent about software licenses is important, however, maybe we should try a friendlier heading? 

Also, the ml5 project has its own license, not MIT! What if we create a new uniform heading something like:

// 👋 Hello! This is an ml5.js made and shared with ❤️.
// Learn more about the ml5.js project: https://ml5js.org/
// ml5.js license and code of conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md

// This example demonstrates _______________
// Any other notes?

_Originally posted by @shiffman in https://github.com/ml5js/ml5-next-gen/pull/66#discussion_r1406387227_

PR #105, opened 2024-03-22
Add neuralNetwork save and load examples

ziyuan-linn:
This PR adds examples for saving and loading neuralNetwork models.

"NeuralNetwork-train-and-save" provides an example of training a model that classifies handPose's keypoint data as "Rock", "Paper", or "Scissors".

"NeuralNetwork-load" provides an example of loading the pre-trained classification model.

Please feel free to leave any suggestions and comments on improving these examples.


examples/NeuralNetwork-load/sketch.js
```diff
@@ -0,0 +1,100 @@
...
+  };
+  classifier = ml5.neuralNetwork(classifierOptions);
+
+  const modelDetails = {
+    model: "model/model.json",
+    metadata: "model/model_meta.json",
+    weights: "model/model.weights.bin",
+  };
+
+  classifier.load(modelDetails, modelLoaded);
```
lindapaiste, line 44: I've got to think about how this can be handled within `preload`.  With the way that I set up the p5 preloading it will work if you include the `modelDetails` as the `modelUrl` property on the `classifierOptions`.  But not if you use the `load` method. 🤔 

I'm playing around with it now and if I move this block into preload I get an error due to not awaiting the `setBackend()` call.  That can be fixed with an `await tf.ready()` in the `neuralNetwork`.

examples/NeuralNetwork-load/sketch.js
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) 2024 ml5
```
shiffman, line 1: Just remembering #76 and we should create a friendlier, less formal standard header for the examples, I don't think a copyright or license reference is needed or helpful, that can be established in this repo.

examples/NeuralNetwork-load/sketch.js
```diff
@@ -0,0 +1,100 @@
...
+  createCanvas(640, 480);
+
+  // Create the webcam video and hide it
+  video = createCapture(VIDEO);
+  video.size(width, height);
+  video.hide();
+
+  // For this example to work across all browsers
+  // "webgl" or "cpu" needs to be set as the backend
+  ml5.setBackend("webgl");
```
shiffman, line 30: Is this needed or can it just work with the default?

examples/NeuralNetwork-load/sketch.js
shiffman: Thank you @lindapaiste it would be great to have this in `preload()` as well!

examples/NeuralNetwork-train-and-save/sketch.js
```diff
@@ -0,0 +1,185 @@
...
+  paperButton = createButton("Add Paper Data");
+  paperButton.mousePressed(addPaperData);
+  scissorsButton = createButton("Add Scissors Data");
+  scissorsButton.mousePressed(addScissorsData);
+  trainButton = createButton("Train and Save Model");
+  trainButton.mousePressed(train);
+
+  // For this example to work across all browsers
+  // "webgl" or "cpu" needs to be set as the backend
+  ml5.setBackend("webgl");
```
shiffman, line 63: I would remove this if the default will work

examples/NeuralNetwork-train-and-save/sketch.js
```diff
@@ -0,0 +1,185 @@
...
+    debug: true,
+  };
+  classifier = ml5.neuralNetwork(classifierOptions);
+
+  // Start the handPose detection
+  handPose.detectStart(video, gotHands);
+}
+
+function draw() {
+  //Display the webcam video
```
shiffman, line 77:
```suggestion
  // Display the webcam video
```

obsessive nitpicking!

shiffman (commented):
These are great thank you @ziyuan-linn, left some nitpicking comments!

examples/NeuralNetwork-load/sketch.js
ziyuan-linn: Ah yes, I created this example by duplicating another example and modifying it and didn't pay attention to the header. Perhaps we can make a separate PR updating all the example headers?

examples/NeuralNetwork-load/sketch.js
ziyuan-linn: The example works without it! However, I do remember discussing this and it is there for some reason, though I don't remember what it is exactly.

shiffman:
Merging!

PR #110, opened 2024-03-23
Move darknet and doodlenet classes to the ImageClassifier folder

lindapaiste:
I think it makes more sense to have these files in the ImageClassifier folder rather than the utils folder.


shiffman:
Thank you @lindapaiste, I agree this is a good change, I'll tag @OrpheasK to review since he worked on this implementation. One note, the "famous" canonical dataset for image classification is [ImageNet](https://www.image-net.org/). This is the dataset I teach about in my intro classes and we often drill into the ml5.js codebase to look at the classes. So I think the file should be named `ImageNet`. `MobileNet` and `DarkNet` are the models that make use of the `ImageNet` labels. 

`DoodleNet` is a model trained specifically for ml5.js by @yining1023 so that file can remain with the same name. I supposed technically the dataset is [`QuickDraw`](https://github.com/googlecreativelab/quickdraw-dataset) and the model is `DoodleNet`  but I this file can keep the same name.

OrpheasK:
I agree with this change as well, makes more sense for these files to be part of the ImageClassifier folder! Regarding the name, I think it would make sense to keep both Imagenet, as it is the original dataset, and Darknet, as it is the only model making use of this specific file the way things are set up, in the classes file name. Maybe `DARKNET_CLASSES_IMAGENET` would be more consistent with `DOODLENET_CLASSES`? (Or even `DOODLENET_CLASSES_QUICKDRAW`? This might be a bit too pedantic but technically easier to understand I guess?) Open to changing it to something else though!

OrpheasK (dismissed):

shiffman:
Thanks for this explanation @OrpheasK! Where does `MobileNet` get the list of classes from? Not this file?

OrpheasK:
Yes, to my understanding the Mobilenet model gets its `topk` classes "built-in" [while being loaded](https://github.com/ml5js/ml5-next-gen/blob/360d14f6e12f7766e9fd5daec3fc0d826f7da86d/src/ImageClassifier/index.js#L101), as its native to tfjs, whereas darknet and doodlenet are external models

shiffman:
I think I was overcomplicating this in the end, I went ahead and just annotated the files with a comment about ImageNet and Quick, Draw. Merging!

PR #111, opened 2024-03-24
Update webpack build configuration

ziyuan-linn:
The PR updates Webpack configuration so it outputs both `ml5.js` and `ml5.min.js` when in production mode. The `main` property in `package.json` is also changed to `dist/ml5.min.js` per the original ml5 library.

Could someone familiar with Webpack take a closer look at this PR since I am not too experienced with it?


shiffman:
Tested and working!

PR #114, opened 2024-03-25
Remove manual fetching of NN `model.json` and `weights.bin` files and use `tf.io.http` instead

lindapaiste:
Ref #113
> * Figure out how to pass fetched data to TFJS using the `IOHandler` interface without [converting to a `File` object](https://github.com/ml5js/ml5-next-gen/blob/360d14f6e12f7766e9fd5daec3fc0d826f7da86d/src/NeuralNetwork/NeuralNetwork.js#L238-L255).

Ref https://github.com/ml5js/ml5-next-gen/pull/84#discussion_r1511567090
> TBH this entire section of the code feels a bit odd and unnecessary. We are fetching the files, creating `File` objects, and then passing them to the TF model with `tf.io.browserFiles`. But TF is definitely capable of fetching files on its own so it seems like we shouldn't have to do that on our end. It seems like we should just be able to pass the URLs. That's a discussion for another time though, because maybe there's a reason that this is necessary and I'm just not seeing it.

Changes:
- Remove manual fetching of files via `axios.get` and conversion to `File` objects.
- Change `tf.io.browserFiles` to `tf.io.http`.
- Set the [`weightUrlConverter` option](https://github.com/tensorflow/tfjs/blob/f0f981fe306bf548e300536aca485c0ffdd6619e/tfjs-core/src/io/types.ts#L531-L541) to use the path from the user's object instead of the path that TFJS finds in `weightsManifest` section of the `model.json` file.

Note: `weightUrlConverter` is specified as an `async` function but I wrote it as a normal function.  We can throw in an `async` keyword if there is concern.

I did try it out and it worked.

image

I even modified the path in the `model.json` file (to `"paths":["./not-the-right-path/model.weights.bin"]`) because in this example the override URL is the same as what's in the `weightsManifest`.  We do this in order to support the p5.js Web Editor, where the actual file path is a remote Amazon S3 URL.


shiffman:
Thank you @lindapaiste this seems like a much more concise way to load files, I tested all the examples and am going ahead to merge.

PR #127, opened 2024-04-22
Make `flipped` an alias for `flipHorizontal`

ziyuan-linn:
This PR adds `flipped` as an alias for `flipHorizontal`.
- Adds alias feature to `handleOptions` function. An `alias` property can now be provided to the `moldObject`, which sets an alias name for an option.
- Changes the mold objects of `handPose`, `faceMesh`, `bodyPose`, and `bodySegmentation` to include `flipped` as an alias for `flipHorizontal`.

TODO:
The `MoveNet` model does not support a `flipHorizontal` option natively. For `MoveNet`, we could try to flip the key-points in ml5 when `flipped` or `flipHorizontal` is set to `true`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,7 +71,6 @@ yarn run build
 
 This will create a production version of the library in `/dist` directory.
 
-
 ## Unit Tests
 
 To run the unit tests, run the following command
@@ -195,6 +194,7 @@ The `moldObject` consists of key-value pairs. The key defines the name of an all
 - `type` (required): A string defining the correct type, can be `"number"`, `"enum"`, `"boolean"`, `"string"`, `"object"`, or `"undefined"`.
 - `default` (required): The default value in case the user does not provide a value or provides an erroneous value for the option.
 - `ignore` (optional): A boolean defining whether the key should be ignored. Defaults to false. Useful when set to a dynamically evaluated value (see section below).
+- `alias` (optional): A string defining an alternative name for this option.
 - Specifically for `type: "number"`:
   - `min` (optional): A number defining the minimum value.
   - `max` (optional): A number defining the maximum value.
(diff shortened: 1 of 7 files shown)


shiffman:
This is great @ziyuan-linn! Shall we add it to the examples (or a selection of examples)? I think it might be nice for the examples, for the most part, to flip the video by default?

ziyuan-linn:
I agree that the examples should have flipped videos by default when possible!

shiffman:
Great! I'm going to go ahead and merge this and open an new issue to track adding `flipped` to the examples.

jackbdu:
@ziyuan-linn Thank you for adding this alias! This [works perfectly fine for `faceMesh`](https://editor.p5js.org/jackbdu/sketches/huMK3RdQh) and [`handPose`](https://editor.p5js.org/jackbdu/sketches/hnwY59twu), but [neither `{ flipped: true }` nor `{ flipHorizontal: true }` works for blazePose `bodyPose`](https://editor.p5js.org/jackbdu/sketches/SoEHcYhqr). Can you take a look what might be the issue?

jackbdu:
@ziyuan-linn Thank you! That would be great, especially considering it is supported in all other models!

In the repo you linked to, I found that [mediapipe BlazePose supports `flipHorizontal`](https://github.com/tensorflow/tfjs-models/blob/cd45df42c8a7605da649789aefb5ae3745f32592/pose-detection/src/blazepose_mediapipe/detector.ts), but it is the only one that does. I wonder if the tensorflow team would be willing to implement it as well.

ziyuan-linn:
@jackbdu Thank you for the information!

I made PR #174 to address this issue, along with some other bugs I found in the process 😂. Please feel free to give any suggestions.

Reaching out to the TFJS team might also be a good idea!

PR #130, opened 2024-05-16
fix typo in faceMesh handleOptions object

ziyuan-linn:
fixes a typo in faceMesh's options handling object


shiffman:
Sorry for missing this one a while back @ziyuan-linn, catching up now!

PR #131, opened 2024-05-16
rename score to confidence across all models

ziyuan-linn:
- adds utility function for renaming a key in an object
- rename the key "score" to "confidence" in the prediction output of `bodyPose`, `handPose`, and `sentiment`
  - `faceMesh` and `bodySegmentation` does not have a score/confidence property
  - `neuralNetwork` and `imageClassifier` already use "confidence" as the key

### DIFF
--- a/examples/BodyPose-keypoints/sketch.js
+++ b/examples/BodyPose-keypoints/sketch.js
@@ -37,7 +37,7 @@ function draw() {
     for (let j = 0; j < pose.keypoints.length; j++) {
       let keypoint = pose.keypoints[j];
       // Only draw a circle if the keypoint's confidence is bigger than 0.1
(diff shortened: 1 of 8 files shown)


shiffman:
Looks great, thank you!!

alanvww:
Hi Peter, just a quick question about the `renameScoreToConfidence()` function, I noticed that this function is not marked as `private` in your comment for these models. Should it be included in the documentation?

And just to be sure, since this change has not been released to the alpha channel yet, will all the `score` be replaced with `confidence` in the 1.0?

ziyuan-linn:
Hi @alanvww, this should be a private method. I forgot to mark it `private`. Thank you for letting me know!

Yes, all the `score` will be replaced with `confidence` for the end user in 1.0.

PR #132, opened 2024-05-25
P5 upload script

ziyuan-linn:
This PR adds a script that uploads the example sketches to the p5 web editor. It uses the same API endpoints as the browser p5 web editor.

Currently, this script can only upload text files. It will log the non-text files in the console, which will require manual uploading. I will look into uploading non-text files in the future, but it is not as straightforward as uploading text files.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,2 @@
+P5_USERNAME=ml5
+P5_PASSWORD=********
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,6 +124,26 @@ npm publish --tag alpha --access public
    https://unpkg.com/ml5@[version]/dist/ml5.js
 ```
 
+## Update p5 Web Editor Sketches
+
+To update the p5 Web Editor sketches, first create a `.env` file with the following content:
+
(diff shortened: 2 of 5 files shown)


shiffman:
This is great @ziyuan-linn! How does this affect all of the ml5.js examples that are linked on the website? Does it also rebuild any "collections"?

ziyuan-linn:
@shiffman Thank you for pointing out the issue! I made some changes to the script. Now it will simply update the sketches instead of deleting and reuploading them. The collections will not be affected either. 

The script compares the directory names in `examples/` to the sketch names on the web editor. If the two names match, the content of the web editor sketch is replaced with the local directory's content without breaking the URL. I added a more detailed explanation in CONTRIBUTING.md.

Please let me know your thoughts!

shiffman:
Thank you @ziyuan-linn, this looks great to me! I haven't tested it but I think I should go ahead and merge. We should discuss how to coordinate with the website team (cc @MOQN @QuinnHe @myrahsa @alanvww) so that we can understand how this affects / works with the documentation that links to the examples.

PR #109, opened 2024-03-23
Create `/tests` folder, remove `canvas` polyfill

lindapaiste:
Fixes #97 
Implements @gohai's recommendations in https://github.com/ml5js/ml5-next-gen/pull/82#issuecomment-2010919727

Changes: 
- Move all test-related code to a new folder `/tests`.
  - Move the `/coverage` folder.
  - Move the jest configuration file (`jest.config.js`).
    - Specify the location in `package.json` since it is no longer at the root.
    - Adjust the paths in the config using `<rootDir>` syntax.
  - Use jest only on the `/unit` folder.
  - Add an `/integration` folder for tests requiring a browser.  These are not run at the moment and will need to be set up in a future PR.
- Uninstall `canvas` package - fixes #97 
  - Remove the polyfill from the `setupTests.js` file. 
  - Move files which involve reading an image into the `/integration` folder.
- Delete Babel config files and put the config in `package.json`.

PR #133, opened 2024-06-04
Remove @tensorflow/tfjs-node from dev dependencies

gohai:
This is causing issues during install on machines (that don't have Xcode installed?). As @lindapaiste wrote in #109: We could consider (..) running the tests using the main/browser version. We will get warnings in the console about how we should be using the Node version, but I get a lot of those still even when I try to use the Node version.


shiffman (commented):
I've tested and everything seems good, merging!

PR #135, opened 2024-06-13
standardize model factory functions

ziyuan-linn:
Currently, some models do not accept the underlying model parameter. `sentiment` also does not allow parameters to be passed in arbitrary order. This PR standardizes the parameters of the factory functions across all models. 

The standized API:

```js
const myModel = ml5.modelName("underlyingModelName", optionsObject, callback);
```
Each of the three parameters is optional, and they can be in an arbitrary order.


src/HandPose/index.js
```diff
@@ -58,6 +60,13 @@ class HandPose {
    * @private
    */
   async loadModel() {
+    this.modelName = handleModelName(
+      this.modelName,
+      ["MediaPipeHands"],
+      "MediaPipeHands",
+      "handPose"
+    );
+
```
lindapaiste, line 69: Personal preference but I feel like it's weird to set `this.modelName` to one value (the raw string) and then replace it with another value (the validated string).  I would move the `handleModelName` call into the constructor.  That way `this.modelName` is always the validated name and there is no confusion about what value it represents.

shiffman:
Apologies @ziyuan-linn I merged #136 first which introduced some conflicts, can you resolve? Thank you for the comments @lindapaiste!

src/HandPose/index.js
ziyuan-linn: Thank you @lindapaiste. This makes sense, I will update it.

ziyuan-linn:
@shiffman This should be good to merge!

shiffman:
Great! Tested examples and merging!

PR #136, opened 2024-06-13
update sentiment

ziyuan-linn:
This PR updates sentiment with the current API standard.
- add support for `webgpu` backend, changing `Sentiment.predict` to an async method
- implement argument-handling for `predict` method
  - add a callback argument to `predict` method
  - add a warning if predict is called without a text argument
  - allow arguments to be passed in arbitrary order
- update the example sketch in accordance with the above changes

### DIFF
--- a/examples/Sentiment/sketch.js
+++ b/examples/Sentiment/sketch.js
@@ -33,15 +33,17 @@ function getSentiment() {
   let text = inputBox.value();
 
   // make the prediction
-  let prediction = sentiment.predict(text);
-
-  // display sentiment result on html page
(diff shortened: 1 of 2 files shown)


lindapaiste:
I'm not sure what convention we decided on for error handling when using callback functions.  In this case the error would be thrown and `gotResults` would not be called.  Which is probably fine?  It's very confusing the user if we call `gotResults(null, error)` because their sketch is still going to crash, but instead of the actual error it would crash with "TypeError: Cannot read properties of null" on the line with `prediction.confidence`.

Issue #134, opened 2024-06-11
Making tfjs runtime the default


ziyuan-linn:
After discussing with @shiffman, I am creating an issue to start a discussion about potentially switching our default runtime to `tfjs` instead of `mediapipe`.

The MediaPipe runtime has been causing some issues:

- **Unable to load two models at the same time**: #69 
- **Memory leak**: #129 

The TFJS runtime, although slightly worse in performance, seems to be more stable and without the above issues. I think it might be a good idea to set the `tfjs` runtime as the default for all models and have `mediapipe` as a more "experimental" option.

Please share your thoughts on this @gohai @MOQN!

gohai:
I too would pick stability over performance, if the better-performing avenue is having those other drawbacks that impact the experience of folks using it.

Just to share some thoughts on the issues @ziyuan-linn mentioned (in case we want to explore addressing those still):

* **Unable to load two models at the same time**: I am wondering if we could work around this with a synchronization mechanism, where the parent `ml5` object keeps a promise (or an array of promises) that get fulfilled whenever a model finishes loading? And a helper function that waits for no outstanding promises before returning? This function could be sprinkled in to the classes that need it?

* **Memory leak**: those seem to be notoriously hard to get to the bottom of... I'd do the same approach as you already did - checking if this also occurs in tfjs example codes, and filing bugs against them if yes, and seeing if there is anything we're doing differently if not

shiffman:
Thank you for the thoughts @gohai! @ziyuan-linn let's go ahead and switch over to `tfjs` for the upcoming soft launch! We can keep issues open and revisit in the future.

PR #140, opened 2024-06-18
switch default runtime to `tfjs`

ziyuan-linn:
This PR changes the default runtime of `handPose`, `bodySegmentation`, `faceMesh`, and `bodyPose` to `tfjs` as discussed in #134.

### DIFF
--- a/src/BodyPose/index.js
+++ b/src/BodyPose/index.js
@@ -93,7 +93,7 @@ class BodyPose {
           runtime: {
             type: "enum",
             enums: ["mediapipe", "tfjs"],
-            default: "mediapipe",
+            default: "tfjs",
           },
           enableSmoothing: {
             type: "boolean",
--- a/src/BodySegmentation/index.js
+++ b/src/BodySegmentation/index.js
@@ -141,7 +141,7 @@ class BodySegmentation {
           runtime: {
             type: "enum",
             enums: ["mediapipe, tfjs"],
-            default: "mediapipe",
+            default: "tfjs",
           },
           modelType: {
             type: "enum",
--- a/src/FaceMesh/index.js
+++ b/src/FaceMesh/index.js
@@ -65,7 +65,7 @@ class FaceMesh {
         runtime: {
           type: "enum",
           enums: ["mediapipe", "tfjs"],
-          default: "mediapipe",
+          default: "tfjs",
         },
         maxFaces: {
           type: "number",
--- a/src/HandPose/index.js
+++ b/src/HandPose/index.js
@@ -73,7 +73,7 @@ class HandPose {
         runtime: {
           type: "enum",
           enums: ["mediapipe", "tfjs"],
-          default: "mediapipe",
+          default: "tfjs",
         },
         modelType: {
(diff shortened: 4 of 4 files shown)


shiffman:
Everything looks good, however, I may have found a small issue with the `BodyPose` model? The mediapipe runtime performs smoothly:

pose-mediapipe

But when I switch to this branch, there is an additional jitter / gltichiness to the results:

pose-tfjs

This is on Mac OS with chrome. Anyone else seeing this?

(videos recorded at SFO ✈️ 😂)

shiffman:
Oh, the above is with blazepose, I'm not seeing the same issue with MoveNet actually (I don't think?). I'll have to test this more thoroughly later!

ziyuan-linn:
I remember looking into this a while ago. The `enableSmoothing` option is `true` by default for `bodyPose`. However, the smoothing algorithm is not implemented on the `tfjs` runtime.

Perhaps we can implement it within ml5? I think `mediapipe` runtime uses a temporal smoothing algorithm, which should be fairly simple to implement.

shiffman:
> I remember looking into this a while ago. The `enableSmoothing` option is `true` by default for `bodyPose`. However, the smoothing algorithm is not implemented on the `tfjs` runtime.
> 
> Perhaps we can implement it within ml5? I think `mediapipe` runtime uses a temporal smoothing algorithm, which should be fairly simple to implement.

Ah, ok, that makes sense! I think we could leave this to later especially since if someone needs it they could manually switch the runtime to mediapipe, is that correct? We should find a place to document this somewhere and perhaps open a new issue for future work / investigation? I don't think this needs to be completed before the upcoming soft launch.

Also, the MoveNet model performs quite stably even without the smoothing.

Hi webteam @QuinnHe @MOQN @myrahsa @alanvww regarding how to document the runtime choices and info about smoothing? Or perhaps this is something we put on a temporary wiki page and link from the documentation as it is not core to the beginner experience?

QuinnHe:
> > I remember looking into this a while ago. The `enableSmoothing` option is `true` by default for `bodyPose`. However, the smoothing algorithm is not implemented on the `tfjs` runtime.
> > Perhaps we can implement it within ml5? I think `mediapipe` runtime uses a temporal smoothing algorithm, which should be fairly simple to implement.
> 
> Ah, ok, that makes sense! I think we could leave this to later especially since if someone needs it they could manually switch the runtime to mediapipe, is that correct? We should find a place to document this somewhere and perhaps open a new issue for future work / investigation? I don't think this needs to be completed before the upcoming soft launch.
> 
> Also, the MoveNet model performs quite stably even without the smoothing.
> 
> Hi webteam @QuinnHe @MOQN @myrahsa @alanvww regarding how to document the runtime choices and info about smoothing? Or perhaps this is something we put on a temporary wiki page and link from the documentation as it is not core to the beginner experience?

Yes, we will take care of it after the soft launch. Just throw an idea for our future reference - maybe add `runtime` in ml5 glossary, and add a lightbulb tip within `Methods` of model pages to link to the ml5 glossary page.

Issue #128, opened 2024-04-25
Incorporating "flipped" property


shiffman:
Picking up on #127, we can now add the `flipped` property to a selection of examples that work with video input. This is a great starter issue for a new contributor!

shiffman:
Hi @sharellb, it occured to me after our meeting today that this could another thing to work on as you get started with learning about ml5.js development! For this issue, I think it's just a matter of going through the examples and incorporating the `flipped` property into some (or all?) of them. After that you could do a review of the docs and propose to the web team where we might add infromation about the `flipped` property (and note how it only works updated version of p5.js, I forget which number exactly!). 

Also there is more info in the p5.js discussion! https://github.com/processing/p5.js/issues/6441

@ziyuan-linn anything I might be forgetting about this new feature of ml5?

ziyuan-linn:
I am for the idea of incorporating `flipped` in all examples! I think for most use cases, the more natural mirror footage would be preferred.

gohai:
I've been coming across "flipping videos" in my fall teaching recently. Curious if I understand it correctly: we ought to `flip` the video element when it gets created **and** pass a flag to the model to flip the input as well?

ziyuan-linn:
Yes, two flags are needed. p5's `flipped` option flips the appearance of the video when the video gets drawn with p5 functions like `image`. The video source object (what gets passed to ml5) is untouched, so a `flipped` flag also has to be passed to the ml5 model to flip the prediction result.

gohai:
Thank you for your insight @ziyuan-linn! I got it to work. The way `flipped` in p5 works does feel a little bit counterintuitive (one would think that flipping the input, once, is all that's needed) - but that's outside our purview

PR #141, opened 2024-06-18
Fix imageClassifier webgpu bug + add teachable machine image example

ziyuan-linn:
This PR changes `dataSync` to `data` in the `classifyInternal` method in order to support `webgpu` backend. `setBackend("webgl")` is no longer needed. The end user API is not affected by this change.

This PR also adds [a teachable machine image example](https://github.com/ml5js/ml5-website-v02-docsify/blob/main/docs/reference/iframes/image-teachable-machine/script.js) as requested by @QuinnHe in ml5js/ml5-website-v02-docsify/issues/96. The only things I changed are removing `setBackend("webgl")` line and changing the color of the text for better visibility. This example should be live on the web editor once we merge this PR and run the p5 upload script.

### DIFF
--- a/examples/ImageClassifier-teachable-machine/index.html
+++ b/examples/ImageClassifier-teachable-machine/index.html
@@ -1,0 +1,22 @@
+<!--
+  👋 Hello! This is an ml5.js example made and shared with ❤️.
+  Learn more about the ml5.js project: https://ml5js.org/
+  ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
+
+  This example demonstrates detecting objects in a live video through ml5.imageClassifier + Teachable Machine.
+-->
+
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>ml5.js imageClassifier + Teachable Machine Example</title>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.9.2/p5.min.js"></script>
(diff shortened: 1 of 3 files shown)


shiffman:
This looks great! Picking up from #128, since Teachable Machine flips the webcam image during training let's include the flipped property in the example. Would you like to add this in before merging @ziyuan-linn? @sharellb you can use this as a model for the other examples if you want to work on those!

QuinnHe:
Thanks Peter! This is awesome!

ziyuan-linn:
@shiffman @sharellb I included the `flipped` property for this example. For this example, only the p5 video had to be flipped. ImageClassifer doesn't deal with landmark points, so it does not need a `flipped` property. For a landmark model like `handPose`, just also add the `flipped: true` to its options object, and the landmark points should be flipped.

shiffman:
> @shiffman @sharellb I included the `flipped` property for this example. For this example, only the p5 video had to be flipped. ImageClassifer doesn't deal with landmark points, so it does not need a `flipped` property. For a landmark model like `handPose`, just also add the `flipped: true` to its options object, and the landmark points should be flipped.

Ah, I see! If the p5.js `flipped` property is enabled, is the mirrored image actually passed to the model? Or would adding the `flipped` property additionally in ml5 enable that? While it will likely work regardless it would be good to be consistent in terms of what is visually displayed and what the model sees. The TM model will likely have been trained with flipped images so it may also be a bit more accurate by passing in flipped images.

ziyuan-linn:
@shiffman I did some quick testing and I think teachable machine actually trains on the original unmirrored webcam footage as well.

I trained a model with mirror camera enabled on teachable machine.
Screenshot 2024-06-19 152925

The same model loaded in to ml5 correctly classifies the underlying unmirrored video.
Screenshot 2024-06-19 153101

We probably don't have to change anything.

shiffman:
Oh, fantastic, thank you the thorough testing! Merging!

PR #143, opened 2024-06-19
Update example sketch names

ziyuan-linn:
This PR prepares for running the p5 upload script

Changes:
- Rename example files.
  - change model names from pascal case to camel case, i.e. `FaceMesh-keypoints` --> `faceMesh-keypoints`.
  - Rephrase some example names for better clarity.
- Update the p5 upload script to match the p5 web editor sketch name and the example directory name in the repo case-insensitively, since most OS have case-insensitive file systems.

Please let me know if any example sketches should be renamed! Once we settle on the sketch names, we can rename the sketches on the web editor to match the directory names in the repo. After that, the p5 upload script should be able to update the sketches in the p5 web editor without breaking any URLs.


ziyuan-linn:
@shiffman Yes that makes sense. changing the names now.

shiffman:
Amazing, here we go!!

Issue #107, opened 2024-03-23
`faceMesh` lips missing `keypoint` 291


jackbdu:
Hi all, I noticed in the [faceMesh parts example](https://editor.p5js.org/ml5/sketches/9y9W7eAee), there is a missing keypoint (291 on the [mesh_map](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/mesh_map.jpg)). See [this sketch with keypoint 291 drawn in red](https://editor.p5js.org/jackbdu/sketches/Js50jtdVg). The same issue is present in [the original tfjs demo](https://storage.googleapis.com/tfjs-models/demos/face-landmarks-detection/index.html?model=mediapipe_face_mesh) so I'm not sure what we should do. Would it make sense to manually add this keypoint in ml5 for the time being (until it's fixed in tfjs)?

ziyuan-linn:
@jackbdu thank you for submitting this issue! We certainly should open an issue with tfjs about this. In the meantime, it should be simple to fix this within ml5 temporarily.

shiffman:
Thank you @jackbdu for reporting and @ziyuan-linn for adding support! How does this affect the documentation? cc @B2xx who is working on this. Does anyone want to volunteer to open the tf.js issue?

jackbdu:
@ziyuan-linn Thank you for addressing this so quickly!

I went ahead and [opened an issue about this on tfjs repo](https://github.com/tensorflow/tfjs/issues/8221).

Issue #67, opened 2023-11-18
FaceMesh faceOval points not in order around the face


dano1234:
Here is a sketch with old ML5 facemesh making a mask
https://editor.p5js.org/dano/sketches/2ULaz2E0v

Here is a sketch with  new ML5 facemesh making a mask
https://editor.p5js.org/dano/sketches/-3gKqxIZD

It appears that all the points for the "faceOval" property come in but they are not in a consecutive order around the face.

As ever thanks for this amazing library

ziyuan-linn:
Sounds great! I am happy to look into this. 

I think the original facemesh came with an annotations object but the new one didn't, so we are now manually creating the annotations within ml5. Apparently, we forgot to take the order of the vertices into account. I will take a look at the old tfjs facemesh library to see how they generate the annotations object.

In addition to using a vertex-sorting algorithm, we could also consider manually creating a map so we can sort the vertices in linear time.

gohai:
Thanks for the report @dano1234! This didn't come up, wasn't attempted, in the summer.

Turns out that MediaPipe, the new underlying model, returns points in a particular order, but for creating contours, we can't simply pair them up sequentially. [Here is how](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/demos/shared/util.js#L207C22-L207C75) Google's own implementation gets this done. And [this](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/src/constants.ts) is the source of these pairings. We should amend `addKeypoints` to take care of this for the contours it creates.

@ziyuan-linn @shiffman I'd be happy to address this Saturday morning (unless Peter already fixed this while I am writing this message 😅)

gohai:
@dano1234 Should be fixed in upcoming test builds! (Here's a [preview](https://editor.p5js.org/gohai/sketches/UzCYkm9Zr) using a local ml5.js.)

We did change the API since the initial build: the points moved to `.keypoints`, so that every contour also has an (easier to use) `.centerX`, `.centerY` properties etc..

PR #144, opened 2024-06-19
add p5 example sketches

ziyuan-linn:
This PR adds a few faceMesh example sketches provided by @MOQN. These examples should help break down the overwhelming amount of keypoints faceMesh provides. These examples should be live on the web editor after merging this PR and running the p5 upload script.

I tested the examples and they all work great! However, the custom shapes example does not seem to draw the lips correctly. 

Screenshot 2024-06-18 220851

I haven't done a closer investigation, but I wonder if the problem is related to https://github.com/ml5js/ml5-next-gen/issues/107.

### DIFF
--- a/examples/faceMesh-bounding-box/index.html
+++ b/examples/faceMesh-bounding-box/index.html
@@ -1,0 +1,22 @@
+<!--
+  👋 Hello! This is an ml5.js example made and shared with ❤️.
+  Learn more about the ml5.js project: https://ml5js.org/
+  ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
+
+  This example demonstrates bounding box facial tracking on live video through ml5.faceMesh.
+-->
+
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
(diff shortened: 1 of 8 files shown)


gohai:
@ziyuan-linn I believe we encountered the lips issue before in #67 and https://github.com/ml5js/ml5-next-gen/commit/5734120ffcc541e5c5d8b463fe7205b288d50745.

Google's tf.js-models code has a [list of point-pairs](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/src/constants.ts#L135) between which one ought to draw a line to arrive at the desired contour.
Our data structure isn't so sophisticated - we try to do a good job now returning the points a meaningful order, but we we can't express e.g. interruptions of the line.

If this is necessary one could either switch to an array of an array of points (`[[{x: 1, y: 2}, ... ], [{x: 2, y: 3}, ... ], ... ]`) - or to a flat array of line segments (`[{x1: 1, y1: 1, x2: 2, y2: 2}, ...]`).

shiffman:
These examples are great!

shiffman:
Merging, we can address the custom shapes example separately, feel free to open an issue for tracking @ziyuan-linn!

PR #145, opened 2024-06-19
Fix badges in README.md

gohai:
Not familiar with the dynamic shields badge formatting, this appears to get rid of the `ALL-CONTRIBUTORS-BADGE` meta-tags to make it work in a single line


ziyuan-linn (approved):
Looks good to me! The badges are working again!

PR #146, opened 2024-06-19
add ml5 version variable

ziyuan-linn:
This PR adds the `ml5.version` variable.


shiffman:
Yay, thank you @ziyuan-linn!

PR #149, opened 2024-06-20
Stylistic tweaks to example codes

gohai:
This is largely about normalizing the comment style (space after slashes, capitalization) *within individual files* - since those instantly jump at the reader. This is also bringing some tweaks made to the docsify MD files back to the examples (ImageClassfier).


shiffman:
Thank you @gohai, this is wonderful!

PR #150, opened 2024-06-20
Update p5 version of example sketches

ziyuan-linn:
This PR updates the example sketches' p5 version to `1.9.4`. I also created a script that automatically updates the p5 versions in examples.


shiffman:
This is great! Nice job with the update script!

PR #151, opened 2024-06-20
rename bodySegmentation example

ziyuan-linn:
The PR renames an example sketch missed earlier.


shiffman:
Whoops, apologies @ziyuan-linn, maybe I'm merging these in the wrong order but looks like a conflict was introduced!

PR #154, opened 2024-06-22
BodySegmentation: Change enum to PERSON

gohai:
I realized that the possible mask types are "person", "background", "parts" - so imho it'd make more sense to go with the same word as the constant for pixels that contain a body/person.


shiffman:
I'm chaotically merging things woohoo!!

PR #174, opened 2024-07-18
Fix body pose bugs

ziyuan-linn:
This PR fixes various bugs in the bodyPose model.
- Rename the confidence value for the entire pose from `score` to `confidence`.
- Fix the [`flipped` and `flipHorizontal` not working issue](https://github.com/ml5js/ml5-next-gen/pull/127#issuecomment-2236929386) by implementing mirroring on ml5's end.
- Fix the keypoints not resizing issue on `MoveNet` model.
  - It seems like `MoveNet` model's keypoints are scaled on the video's intrinsic size ("videoWidth" and "videoHeight") instead of display size ("width" and "height"). This contradicts the behavior of `handPose`, `faceMesh`, and the `BlazePose` model. This fix rescales the keypoints to the video display size.

### DIFF
--- a/src/BodyPose/index.js
+++ b/src/BodyPose/index.js
@@ -208,6 +208,17 @@ class BodyPose {
           modelConfig.modelType =
             poseDetection.movenet.modelType.MULTIPOSE_LIGHTNING;
       }
+      this.runtimeConfig = handleOptions(
+        this.config,
+        {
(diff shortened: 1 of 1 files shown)


shiffman:
Looks great! This change might be worth doing a patch release to practice and develop a process for documenting releases? cc @sharellb @alanvww

PR #176, opened 2024-07-22
add bounding box mirroring

ziyuan-linn:
This PR implements mirroring for `MoveNet`'s bounding box. When `flipped` is set to `true`, the bounding box will be horizontally flipped.


shiffman:

Issue #175, opened 2024-07-22
`bodyPose` and `bodySegmentation` failing to preload in p5 instance mode


jackbdu:
Creating a new issue to track the progress of this:

@lindapaiste Thank you for adding support for instance mode. I was able to preload most models (`ml5.faceMesh`, `ml5.handPose`, `ml5.imageClassifier`, `ml5.neuralNetwork`, and `ml5.sentiment`) but I'm getting an error for `ml5.bodyPose`. See [this minimal sketch demonstrating the issue](https://editor.p5js.org/jackbdu/sketches/c4XaQbpev). I also included my temporary fix in the comments, which creates blank `window._incrementPreload()` and `window._decrementPreload()` to avoid this error.

I'm also getting an error for preloading `ml5.bodySegmentation` in instance mode, and that could be fixed by declaring a global variable `let video;`.

_Originally posted by @jackbdu in https://github.com/ml5js/ml5-next-gen/issues/81#issuecomment-2067780151_

lindapaiste:
We need to remove this call:
https://github.com/ml5js/ml5-next-gen/blob/542e5a5171a428e807451ca901cc130b54d6f8e9/src/BodyPose/index.js#L56
(and any others like it)

There will be no `window._incrementPreload` when using instance mode because `_incrementPreload` will be a property of the p5 instance, rather than a property of the window.

lindapaiste:
> I'm also getting an error for preloading `ml5.bodySegmentation` in instance mode, and that could be fixed by declaring a global variable `let video;`.

Likely we have some code that is using an improperly scoped `video` variable.  We definitely should not have a global/window scoped `video` variable anywhere.  I suspect it's an issue where we are using `this.video` in a context where `this` is actually the window.  That sort of thing can be fixed by binding the function to the correct `this` context.  I would need to dig into the code to find where the bad variable is.

lindapaiste:
We're using a variable `video` that doesn't exist on like 34 of `BodySegmentation`.

https://github.com/ml5js/ml5-next-gen/blob/84419b3d2808aac1a6214c24f1f6b00d918127f2/src/BodySegmentation/index.js#L34

shiffman:
I don't believe this is an issue anymore, I tested the example linked in the original comment with the latest release and there is no error. If I am mistaken I will be glad to re-open. Also, while we will continue to support `preload` we are moving to p5.js 2.0 `async` and `await` model.

PR #178, opened 2024-07-25
BodyPose cleanup

ziyuan-linn:
This PR updates the comments and refactors the code of BodyPose. No change to the user API.
- update incorrect license info: MIT License -> ml5.js License.
- add file description and helpful links at the beginning of the file.
- format and clean up options object JSDoc comment.
- move mold objects/schemas for `handleOptions` next to the options object comments.
- add comment description to each property of `BodyPose` class.
- update various comments for clarity
- refactor some variable names for clarity

This PR also removes the `window._incrementPreload` and `window._decrementPreload` calls to address the bug in #175.

### DIFF
--- a/src/BodyPose/index.js
+++ b/src/BodyPose/index.js
@@ -28,37 +28,160 @@ import { handleModelName } from "../utils/handleOptions";
 import objectRenameKey from "../utils/objectRenameKey";
 import { isVideo } from "../utils/handleArguments";
 
+/**
+ * User provided options object for BodyPose with MoveNet model,
+ * the default model used by ml5.
+ * @typedef {Object} MoveNetOptions
+ * @property {string|undefined} modelType             - The type of MoveNet model to use.
+ * @property {boolean|undefined} enableSmoothing      - Whether to use temporal filter to smooth
+ *                                                      keypoints across frames.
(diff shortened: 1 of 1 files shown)


shiffman (commented):
This looks great, thank you @ziyuan-linn! Looks like you tagged @alanvww for a review, I'm happy to merge afterwards!

alanvww (approved):

alanvww:
Awesome work! Thanks for letting me know about the update, I will work on the doc ASAP.

@shiffman We are ready to merge!

lindapaiste:
Thoughts:

It might make sense to move functions like mirrorBoundingBox to a utility file, if we want to use that logic in multiple models.

There are some types that can be more specific, but it’s annoying if we have to write out the union of possible image types every time.  The return types could also be more specific (an array of what?).  We can define a type for the prediction object and instead of Promise<Array> it can be Promise<Array<BodyPrediction>>.  That might not be worth the effort, unless…

It would be really great if we can figure out a way to sync property descriptions between the code comments and the documentation.

shiffman:
Thank you for the feedback @lindapaiste, feel free to open any issues with proposals for these changes if you like!

Issue #166, opened 2024-07-11
`topk` in the `options` for `imageClassifier` doesn't work


jackbdu:
In `ml5@1.0.1`, when initializing the `imageClassifier` object, [passing `{ topk: 5 }` as `options`](https://editor.p5js.org/jackbdu/sketches/4d7ufw-Lh) does not seem to affect the number of labels to return. On the other hand, specifying `kNumber` works fine when calling `imageClassifier.classifyStart(media, ?kNumber, callback);`

Tagging relevant issues/PRs for reference: #71 #83

OrpheasK:
I looked into this and while there is a simple workaround to correct it and have the `topk` parameter actually apply to the number of classes, maybe it is a chance to address the first issue of [#83](https://github.com/ml5js/ml5-next-gen/issues/83) raised by @lindapaiste an mentioned here. The `topk` parameter can only currently apply to MobileNet and is overridden by `kNumber`.

Since it is correctly documented in the website, correcting the issue (and perhaps renaming the parameter @shiffman) is a simple solution, we might also consider however which parameter to make the universal one for choosing the number of classes.

shiffman:
This has now been addressed in #179, I've tested this with the `imageClassifier-webcam` example, adding the argument to `classifyStart()` produces the desired result:

```js
// Get the top 25 labels
classifier.classifyStart(video, 25, gotResult);
```

Should an example include this feature or is it just a matter of making sure the documentation is clear?

Issue #83, opened 2024-03-03
Discussion: ImageClassifier


lindapaiste:
I'm opening this thread as a place for us to discuss ongoing work on the `ImageClassifier` model.  Some of the tasks are:
 - [ ] Decide where the user should specify the number of classes to return (the `topk` variable).  Should it be when they create the model?  When they call `classify`/`classifyStart`?  Or both?  In the new version of the detector models we eliminated the ability to pass `options` to the `detect` function.  Following that example we would support `ml5.imageClassifier('doodlenet', { topk: 5 })` but not `myClassifier.classify(image, 5)`.  Right now passing the option to `ml5.imageClassifier()` only works for `mobilenet`.
 - [ ] Separate handling which only applies to one model type https://github.com/ml5js/ml5-library/pull/1362/
 - [ ] Figure out where we want to handle conversion to Tensor, related to https://github.com/ml5js/ml5-library/pull/1396
 - [ ] Clean up preprocessing code using the utilities created in https://github.com/ml5js/ml5-library/pull/1398
 - [ ] See if there are parts of the legacy code that don't make sense.  For example we look for a property `ml5Specs` in the `model.json` file for custom models.  That property would only be set if the user trained their model using the ml5 `FeatureExtractor`, which we haven't ported over here.  But users may have trained models on the old version so I guess we should keep it?
 - [ ] Look into what a model from TeachableMachine actually looks like and see if we are handling it properly.  Is the image size _always_ `224`?  Is our range of normalized values correct?  Can we [infer these values from the model](https://github.com/ml5js/ml5-library/blob/b4960e9f8d5673830b6e149ad9da0e3548ec636a/src/utils/tensorInput.js) in case the user has a custom trained model which doesn't match our expectations? 
 - [ ] Write unit tests for all model types which make sure that the model gives the correct prediction.  There's a lot in https://github.com/ml5js/ml5-library/blob/b4960e9f8d5673830b6e149ad9da0e3548ec636a/src/ImageClassifier/index.test.js but even that is incomplete.

jackbdu:
[...]

PR #179, opened 2024-07-27
Image Classifier update

OrpheasK:
This PR fixes the issue #166 that @jackbdu brought into attention and also offers a chance to decide upon for the use/name of the `topk` parameter in conjunction with `kNumber` as already mentioned in issue [#83](https://github.com/ml5js/ml5-next-gen/issues/83) by @lindapaiste. 

Also recently mentioned in #83 is the support of local loading of TM models among other subjects, which can be incorporated in this or another PR depending on the priority each aspect will be given.

### DIFF
--- a/src/ImageClassifier/index.js
+++ b/src/ImageClassifier/index.js
@@ -248,7 +248,7 @@ class ImageClassifier {
       "image",
       "No input image provided. If you want to classify a video, use classifyStart."
     );
-    return callCallback(this.classifyInternal(image, number), callback);
+    return callCallback(this.classifyInternal(image, number || this.topk), callback);
   }
 
   /**
@@ -269,10 +269,9 @@ class ImageClassifier {
     // Function to classify a single frame
     const classifyFrame = async () => {
       await mediaReady(image, true);
-      await this.classifyInternal(image, number);
       // call the callback function
-      callCallback(this.classifyInternal(image, number), callback);
-
+      await callCallback(this.classifyInternal(image, number || this.topk), callback);
+      
       // call recursively for continuous classification
       if (!this.signalStop) {
         requestAnimationFrame(classifyFrame);


ziyuan-linn (approved):
This looks good to me! I don't really have a strong opinion on the naming of `topk`, but perhaps another option could be `classes`. `classes: 4` returns the top 4 classifications. 

Personally I think we should only use the `ml5.imageClassifier({topk: 4})` and not `myImageClassifier.classify(image, 4)`, so the interface is more consistent with the other models. This might be a breaking change though, and I don't feel strongly about this either.

lindapaiste:
`resultCount` might also make sense.

shiffman:
> Personally I think we should only use the ml5.imageClassifier({topk: 4}) and not myImageClassifier.classify(image, 4), so the interface is more consistent with the other models. This might be a breaking change though, and I don't feel strongly about this either.

@ziyuan-linn can you elaborate on this? What other models follow this pattern? Perhaps it's better to conform the other models to how we are doing it here? I think having the # of classes returned specified with `classify()` is more intuitive?

ziyuan-linn:
@shiffman The other pretrained models like `BodySegmentation` and `FaceMesh` only accept a media element and an optional callback on the detection methods and do not accept any kind of options. For example `faceMesh.detect(image, gotFaces);`. All the options are set during the `ml5.faceMesh()` call. 

I think setting the number of classes returned is unique to `ImageClassifier` (and maybe soundclassifier?).

I agree that the number of classes on the `classify` method is more intuitive.

shiffman:
Apologies for letting this PR languish for a bit! Based on the discussion we had at our previous Thursday meeting, we decided to allow for passing in the number of desired classes + confidence scores to be returned as an argument to the `classify()` function itself. We can then extend this pattern to other models that work in a similar way (right now it would only be `ml5.neuralNetwork()` for the `"classification"` task.  I'll open an issue regarding incorporating this in examples and website documentation.

Issue #112, opened 2024-03-24
Discussion: `NeuralNetwork`


lindapaiste:
I've been working a lot on refactoring the NeuralNetwork code so I wanted to have a place to ask questions as they come up.

First up:

`classifyMultiple` and `predictMultiple` are a bit odd because they are [actually identical](https://github.com/ml5js/ml5-next-gen/blob/360d14f6e12f7766e9fd5daec3fc0d826f7da86d/src/NeuralNetwork/index.js#L831-L873) to the `classify` and `predict` functions.  They call the same `classifyInternal`/`predictInternal` function with the same `_input` argument.

We determine whether it's single or multiple by looking at the input.  This could be potentially buggy.  For example if you pass an array of inputs which has length 1 to the `classifyMultiple` method then you get back the same results as you would from a single classification.  Not an array of results (array of arrays) with length 1.  That's because we're determining whether to return a nested array [by looking at the length](https://github.com/ml5js/ml5-next-gen/blob/360d14f6e12f7766e9fd5daec3fc0d826f7da86d/src/NeuralNetwork/index.js#L981-L987) with no knowledge of which method was used.

We have a few options here:
1. Leave things as-is.
2. Have one `classify` method which can handle single or multiple classification (This is how the TFJS model.predict() works, but I don't like this because it is confusing to explain in the docs).
3. Ensure that `classify` always returns a single classification and `classifyMultiple` always returns an array.  We would validate the inputs and log warnings or throw errors if the user provides the wrong type of input.
4. Remove batched multiple classification entirely, and have one `classify` method which can only accept a single input.

PR #181, opened 2024-07-31
reorder metadata legend before results

### LINKED ISSUE #180: `ml5.neuralNetwork` reorders labels that are numerical strings
Hi all,

I've been trying to train a [MNIST](https://yann.lecun.com/exdb/mnist/) handwritten digit classifier with `ml5.neuralNetwork` using numerical strings like `"1"`, `"2"`, `"3"` as labels. The trained neural network seems to classify inputs completely randomly even though the loss is converging.

After hours of debugging, I realized the neural network would behave as expected if I change my labels to be non-numerical strings like `"Digit 1"`, `"Digit 2"`, `"Digit 3"`. I did a few more [tests with the neuralNetwork color classifier example](https://editor.p5js.org/jackbdu/sketches/Y5nVeG8BX) and found out that numerical strings seem to be reordered when added as outputs, hence producing inaccurate classifications after training.

I haven't had a chance to look at the source code, but I suspect this might have something to do with how JavaScript treats numerical strings?

ziyuan-linn:
This PR fixes #180 by sorting `Object.entries(legend)` via the one hot array, ensuring that its ordering matches the ordering of `unformattedResults`.

### DIFF
--- a/src/NeuralNetwork/index.js
+++ b/src/NeuralNetwork/index.js
@@ -1102,7 +1102,12 @@ class DiyNeuralNetwork {
     if (meta !== null) {
       const label = Object.keys(meta.outputs)[0];
       const vals = Object.entries(meta.outputs[label].legend);
-
+      vals.sort((a, b) => {
+        return (
+          b[1].reduce((acc, curr) => acc * 2 + curr, 0) -
+          a[1].reduce((acc, curr) => acc * 2 + curr, 0)
+        );
+      });
       const formattedResults = unformattedResults.map((unformattedResult) => {
         return vals
           .map((item, idx) => {
--- a/src/NeuralNetwork/index.js
+++ b/src/NeuralNetwork/index.js
@@ -1102,12 +1102,14 @@ class DiyNeuralNetwork {
     if (meta !== null) {
       const label = Object.keys(meta.outputs)[0];
       const vals = Object.entries(meta.outputs[label].legend);
+      // sort the legend array results by each item's one-hot encoding
       vals.sort((a, b) => {
         return (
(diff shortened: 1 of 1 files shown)


lindapaiste:
Can you explain what this sort function is doing?  It’s not clear at first glance.

It does feel like this is a roundabout solution.  It’s probably fine for now to just fix the issue but I wonder if there’s a better way to structure the metadata objects to avoid the issue.

ziyuan-linn:
Thank you for pointing out the ambiguity. I also added a comment in the code.

This function reorders the `Object.entries(legend)` array from something like:
```js
[
  ["1", [0, 1, 0, 0],
  ["2", [1, 0, 0, 0],
  ["3", [0, 0, 0, 1],
  ["4", [0, 0, 1, 0],
]
```
to:
```js
[
  ["2", [1, 0, 0, 0],
  ["1", [0, 1, 0, 0],
  ["4", [0, 0, 1, 0], 
  ["3", [0, 0, 0, 1],
]
```

I agree that a better long-term solution is to restructure the metadata object. Perhaps we can track this in a new issue or add something to #112.

shiffman:
@ziyuan-linn I'm going to go ahead and merge this for the "bug fix" but feel free to open an issue or add to an issue's discussion!

lindapaiste:
I think you could get the same sort order with `vals.sort((a, b) => a[1].indexOf(1) - b[1].indexOf(1))`. Personally I find `.indexOf(1)` more clear than `.reduce((acc, curr) => acc * 2 + curr, 0)`, but this is already merged so no big deal.

Issue #182, opened 2024-08-02
BodySegmentation: flipped option doesn't work


manuel-martos:
I've been experimenting a bit with combining the different models, and it seems like the 'flipped' option doesn't work for the bodySegmentation model. See the following screenshot as a reference:

![image](https://github.com/user-attachments/assets/4c5826ad-b945-4515-8b48-ce1b04cc4a05)

ziyuan-linn:
Hi @manuel-martos, thank you for reporting this!

When the `flipped` option is set to `true`, ml5.js sets `flipHorizontal` to true in the config object of the underlying Tensorflow.js models. I looked into the issue and discovered that the `flipHorizontal` option is documented but never implemented in the [underlying tfjs body-segmentation model](https://github.com/tensorflow/tfjs-models/tree/master/body-segmentation).

I am opening a PR to implement the flipping feature in ml5.js.

PR #184, opened 2024-08-02
add fllipImageData function

ziyuan-linn:
This PR fixes the bug where `flipped` and `flipHorizontal` options do not work for `BodySegmentation` (#182).

- adds a function that flips an `ImageData` objects horizontally
- flips segmentation output with the above function when `flipped` or `flipHorizontal` is set in the options object


shiffman:
Thank you @ziyuan-linn!

PR #185, opened 2024-08-06
change upload script ml5 version number

ziyuan-linn:
Instead of specific version numbers like `ml5@1.0.1`, use `ml5@1`.

This way if people make copies of the example scripts, the ml5 version will automatically roll over to the newest non-breaking release.


shiffman:
I'm a little afraid of this (haha) b/c I prefer knowing the *exact* version I'm using but I can see how this is a good idea, let's go for it! (We can always roll back.)

PR #189, opened 2024-08-06
update dev environment setup instruction

ziyuan-linn:
The PR updates the dev environment setup instruction to include a workaround for potential yarn versioning issues. 

cc @sharellb Just FYI I think the issue we ran into might be caused by a different version of yarn installed in another nvm installation of node.js. However, I am not able to consistently reproduce the error. The fix `npm i -g --force yarn` also did not always work. If someone runs into this problem again, I think the best workaround for now is to just uninstall other nvm node.js installations that have the wrong version of yarn.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,6 +89,8 @@ yarn
 yarn start
 ```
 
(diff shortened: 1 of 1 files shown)


shiffman:
Looks good, thanks for all the work on the release @sharellb and @ziyuan-linn!

PR #190, opened 2024-08-07
remove experimental warning in console message

ziyuan-linn:
This PR removes the experimental warning in the console message since we are past `1.0.0`.


shiffman:
woohoo!

PR #195, opened 2024-08-22
Add object detection implementation

douaaz:


src/ObjectDetector/index.js
```diff
@@ -0,0 +1,142 @@
...
+    const argumentObject = handleArguments(...inputs);
+    argumentObject.require("image", "An html or p5.js image, video, or canvas element argument is required for detect().");
+    const { image, callback } = argumentObject;
+
+    await mediaReady(image, false);
+
+    const predictions = await this.model.detect(image);
+    console.log('raw result from cocoSsd', predictions);
+
+    const result = predictions;
```
gohai, line 76: Perhaps worth having a look at the previous version of ml5 to see in which format it returned data to the user. Not that you necessarily need to imitate it, but not so sure about `bbox`..!

From a quick look, previously the result was an array of objects, each with the following properties:
```
label: String,
confidence: Number,
x: Number (px),
y: Number (px),
width: Number (px),
height: Number (px),
normalized: {
  x: Number (0-1),
  y: Number (0-1),
  width: Number (0-1),
  height: Number (0-1),
}
```

(the `normalized` part is perhaps less common in ml5, and perhaps we could drop this - but the rest seems to be similar to e.g. `faceMesh` returns results presently)

src/ObjectDetector/index.js
```diff
@@ -0,0 +1,142 @@
...
+   */
+  async detect(...inputs) {
+    const argumentObject = handleArguments(...inputs);
+    argumentObject.require("image", "An html or p5.js image, video, or canvas element argument is required for detect().");
+    const { image, callback } = argumentObject;
+
+    await mediaReady(image, false);
+
+    const predictions = await this.model.detect(image);
+    console.log('raw result from cocoSsd', predictions);
```
gohai, line 74: Great for debugging, but normally good to comment this out when sharing this with others (as people are often time confused by console messages that they didn't cause themselves, especially when they are frequent).

shiffman:
This is so exciting to see, if it's ok, I will leave a few comments as well!

examples/objectDetector-dimLight/sketch.js
shiffman, line 1: This is a lovely implementation, but I wonder if it's outside the scope of a basic set of Object Detection examples since it's not more about "image processing" than the model itself. Perhaps a separate tutorial could be written for the community page about pre-processing images since it could be applied across many models, not just object detection.

Regardless, if we were to keep this example I would suggest rewriting it with p5.js functions rather than the native JS code which will be unfamiliar to beginners.

examples/objectDetector-single-image copy/objects.jpg
shiffman, line 1: Generally speaking, I think it's preferable to demonstrate examples without people in them. Who is this person? Do we have permission to use their face? Perhaps there is an image that can demonstrate a wider variety of classes inside the model.

examples/objectDetector-single-image copy/sketch.js
```diff
@@ -0,0 +1,73 @@
...
+  // Draw the image
+  image(img, 0, 0);
+
+  // Loop through all the detected objects and draw bounding boxes with labels
+  for (let i = 0; i < objects.length; i++) {
+    let object = objects[i];
+    let x = object.bbox[0];
+    let y = object.bbox[1];
+    let w = object.bbox[2];
+    let h = object.bbox[3];
```
shiffman, line 43: Do we want to consider renaming the underlying properties that come out of the model, something like the following might be more intuitive:

```js
    let x = object.bbox.x
    let y = object.bbox.y;
    let w = object.bbox.w;
    let h = object.bbox.h;
 ```
 
 This also offers the opportunity for "destructuring" though this is perhaps a concept less familiar to beginners:
 
 ```js
let { x, y, w, h } = object.bbox;
 ```

File: examples/objectDetector-single-image copy/sketch.js [hunk 4]
```diff
@@ -0,0 +1,73 @@
...
+    fill(object.color.r, object.color.g, object.color.b);
+    textSize(16);
+    text(object.class, x + 5, y + 15);
+  }
+}
+
+// Callback function for when objectDetector outputs data
+function gotObjects(results) {
+  // Save the output to the objects variable and assign a random color to each object
+  objects = results.map(object => {
```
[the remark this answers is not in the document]
shiffman: I think this also might confuse the simplicity of this example, I might just use a default color for all objects and not add properties to the results coming from the model.

src/ObjectDetector/index.js
```diff
@@ -0,0 +1,158 @@
...
+    const argumentObject = handleArguments(...inputs);
+    argumentObject.require(
+      "image",
+      "An html or p5.js image, video, or canvas element argument is required for detect()."
+    );
+    const { image, callback } = argumentObject;
+
+   // await mediaReady(image, false);
+
+   // const predictions = await this.model.detect(image);
```
gohai, line 81: The two lines above look like they were commented-out by mistake?

enkatsu:
Hi all, thank you for the great work on this PR!
I just wanted to kindly ask — what’s the current status of this pull request?
I'm really looking forward to seeing this object detection feature merged, and if there's anything I can do to help move it forward (resolving conflicts, testing, etc.), I’d be more than happy to contribute.
Apologies if this comes across as rushing — I truly appreciate the work being done here!
Thanks again 🙏

shiffman:
Hi @enkatsu thanks for question and your interest and enthusiasm! This is on the agenda for summer research and we hope to have something complete by the fall term, stay tuned! The summer work on ml5.js will begin at the start of June so check back then!

enkatsu:
Hi @shiffman, thanks for the update! That’s great to hear — I’m excited to see how things develop over the summer.
Appreciate all your hard work!

PR #200, opened 2024-09-09
Add offline module

ziyuan-linn:
The PR implements an offline version of the ml5.js library. 

Webpack will now build `ml5-offline.js` and `ml5-offline.min.js` in addition to `ml5.js` and `ml5.min.js`. The file `ml5-offline.js` and `ml5-offline.min.js` is self-contained and will not make additional network requests. Once a user downloads `ml5-offline.min.js`, the ml5 library can run entirely without internet.

The entry point for the offline model is `src/offline/index.js`. It imports the ml5 object, injects a `loadOfflineModel` function into each ml5 model's prototype, and exports the ml5 object with the modified models. The approach keeps the offline implementation mostly separate from the main source code.

Changes:
- add a script that downloads the pre-trained models from Kaggle.com.
  - The script converts the model JSON and binary files to JS files.
  - `model.json` is written to a JS file.
  - binary files converted to `Uint8Array` and then written to JS files.
  - This is done so that Webpack can pack the model files and source code into a single file. 
- implement a `loadOfflineModel` function.
  - This function loads the above data into `File` objects and creates object URLs for them.
  - The object URLs are then injected into the "model URL" fields (e.g. `landmarkModelURL`) of the ml5 model.
- For the `ml5-offline.js` build, add the `loadOfflineModel` functions into object prototypes of each ml5 model.
- Call the `loadOfflineModel` function from each ml5 model during initialization if the function is defined.

To test this implementation, find an example sketch and change `ml5.js` to `ml5-offline.js` in `index.html`.

Note: This is WIP, `ml5-offline.js` only works for `HandPose` with `modelType: "lite"`.


shiffman:
This is exciting to see thank you @ziyuan-linn for working on this! One consideration is that running a stand-along node.js script for downloading the models may involves concepts/steps unfamiliar to beginners especially those used to working only in the p5.js web editor. I think it's ok for this to be the primary way to get the models since this is an "edge" use case, however, we should probably include some detailed documentation for total beginners. I could also record a video tutorial on this topic. Alternatively we could consider providing links to some model downloads directly but this might be hard to maintain.

jackbdu:
@ziyuan-linn Thank you for working on this!! I have been looking forward to the offline support! I think automatically downloading the models would be nice.

I built this PR locally with `yarn build` and it took more than two hours. Was it normal? Activity Monitor showed webpack was using 1.35GB memory and 10% CPU, not sure what the bottleneck was. In any case, `HandPose` examples worked fine with the offline build.

I am trying to add offline support to `FaceMesh` by following your `HandPose` implementation. I modified relevant code in [`src/offline`](https://github.com/ml5js/ml5-next-gen/tree/add-offline-module/src/offline), [`src/FaceMesh`](https://github.com/ml5js/ml5-next-gen/tree/add-offline-module/src/FaceMesh), and [`scripts/fetchModelFiles.js`](https://github.com/ml5js/ml5-next-gen/blob/add-offline-module/scripts/fetchModelFiles.js). I was able to use `fetchModelFiles.js` to download the `FaceMesh` model from `"https://www.kaggle.com/api/v1/models/mediapipe/face-landmarks-detection/tfJs/face-mesh/1/download"`, but the `FaceMesh` example sketches still seem to be requesting the model remotely.

@ziyuan-linn Can you shed some light on this? Let me know if you need more information.

ziyuan-linn:
@jackbdu Thank you for looking into this and sorry for the late response! When I was working on this, the builds complete in minutes. I tried rerunning the build again today and it is taking a long time. It seems like the memory is hitting some sort of limit. My process is sitting at nearly 4 GB of memory with nearly no CPU. I am assuming the heavy memory usage is coming from building the giant `Uint8Array` of model weights into the library itself, but I am not sure how this can be addressed.

For FaceMesh, I think you would need two URLs: one for the landmarks model and one for the detector model. It seem like you have the landmark model and is missing the detector model?

I stopped working on the offline module because after having a discussion with Dan we realized that the offline package will become very bloated with the model weights. I think having documentation on the website about using the ml5 library offline would be a better approach than maintaining an offline version of the library. The current library already have options like `detectorModelUrl` and `landmarkModelUrl` that allows the users to load offline weights. We can point the users to Kaggle to download the weights, or even host a copy of the weights on the ml5 website for download.

However, I understand that there may be benefits to having a offline version of the library that just works out of the box. Let me know what you think!

jackbdu:
@ziyuan-linn Thank you for looking into this and the clarifications! I totally agree with your conclusion! As long as we have clear documentation on how to run the models offline, there is no need for us to maintain an offline version of the library!

Issue #206, opened 2024-09-23
nn.loadData() appears to be broken


gohai:
When loading a JSON file produced by `nn.saveData()`, the following error message is returned (with the latest 1.x version of the library): `data must be a json object containing an array called "data"`

To reproduce: try pressing `l` in [this sketch](https://editor.p5js.org/gohai/sketches/LGnxcG-5C). The `data3.json` file has been produced using the same code (same ml5 version).

ziyuan-linn:
@gohai Thank you for submitting this issue! Looks like there was an incorrect call to `JSON.stringify` on an already stringified JSON object. I opened a PR to fix this.

PR #207, opened 2024-09-23
fix json parse bug in neuralNetworkData loadData

ziyuan-linn:
This PR fixes the bug in #206. 

Changes:
- Run formatted on `NeuralNetwork`
- Fix the bug where `JSON.stringify` was necessarily called
- Update the console message for readability


shiffman:
Thank you for this @ziyuan-linn! I think an older version of ml5.js supported `csv` data as well? So that is something we could consider for the future. But I don't think it's a priority. If anyone feels it's a need or is excited to work on it, we can open an issue.

Issue #204, opened 2024-09-17
`normalizeData()` does not handle only zeros in one of the inputs to the neural net


jchwenger:
Hi there,

I've been noticing what seems to be a little bug.

If you go to [this example](https://editor.p5js.org/ml5/sketches/eGHBdmCLe) and comment out all examples for one class, e.g. all the "blue-ish" objects, training fails with infinite loss, because the normalized data has `NaN`s in the third coordinate of xs (check `classifier.data`). I bumped into this in [this sketch](https://editor.p5js.org/jchwenger/sketches/71cFcKwS2), where again I would like this to work even if only two classes are provided. As long there's nonzero numbers everywhere in the training input, it's fine, but normalizing zeros doesn't seem to work. I wonder where you would want to add a check for this: directly in `normalizeDataRaw`?

ziyuan-linn:
Hi @jchwenger,

Thank you again!

It seems like the code uses `(value - min) / (max - min)` to compute the normalized values. This will divide by zero if `min == max`, which is the source of the `NaN`. When all values of a parameter are the same, the bug will occur.

I opened PR #208 to address this.

From local testing, the linked sketch now works even if not all classes are provided.
<img width="300" alt="Screenshot 2024-09-23 at 6 10 07 PM" src="https://github.com/user-attachments/assets/77ec3052-5d03-4303-a223-4c9800290c05">

jchwenger:
Hi @ziyuan-linn, awesomeness! Nice thinking, this check, looking forward to the pr being integrated!

PR #208, opened 2024-09-23
handle min = max case during data normalization

ziyuan-linn:
This PR addresses issue #204 by adding a check for `min === max` in `normalizeValue`, preventing the divide by zero situation.

### DIFF
--- a/src/NeuralNetwork/NeuralNetworkUtils.js
+++ b/src/NeuralNetwork/NeuralNetworkUtils.js
@@ -11,6 +11,10 @@ class NeuralNetworkUtils {
    */
   // eslint-disable-next-line class-methods-use-this
   normalizeValue(value, min, max) {
+    // When min is equal to max, set everything to 0
+    if (min === max) {
+      return 0;
+    }
     return (value - min) / (max - min);
   }
 
--- a/src/NeuralNetwork/NeuralNetworkUtils.js
+++ b/src/NeuralNetwork/NeuralNetworkUtils.js
@@ -4,12 +4,11 @@ class NeuralNetworkUtils {
   }
 
   /**
-   * normalizeValue
-   * @param {*} value
-   * @param {*} min
-   * @param {*} max
+   * Normalize a value between min and max, return 0 if min === max
+   * @param {number} value - The value to normalize
+   * @param {number} min - The minimum bound
+   * @param {number} max - The maximum bound
    */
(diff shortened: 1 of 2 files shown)


src/NeuralNetwork/NeuralNetworkUtils.js
```diff
@@ -11,6 +11,10 @@ class NeuralNetworkUtils {
    */
   // eslint-disable-next-line class-methods-use-this
   normalizeValue(value, min, max) {
+    // When min is equal to max, set everything to 0
+    if (min === max) {
+      return 0;
```
shiffman, line 16: So just thinking about this a bit, should we log some sort of warning in the console? Perhaps we should also just return `min` (or `max`) to keep the original values? And log that they cannot be normalized?

shiffman (commented):
Just left a small question to think about before merging!

src/NeuralNetwork/NeuralNetworkUtils.js
ziyuan-linn: I think a warning would be helpful! I am not sure how to word it though. Perhaps "The input parameter <x> cannot be normalized because all values for that parameter are identical. The NeuralNetwork can still be trained, but please check your input data".

If we keep the original (non-zero) data, that parameter effectively becomes a bias value for the next layer since it is constant. If we zero that parameter we are pretty much deleting it from the neural network. I'm not sure whether keeping it could throw off the training, but I'm leaning toward just zeroing the parameter since it doesn't contribute to training anyway.

shiffman: Ah, I see what you mean! How about:

```js
console.warn("Normalization failed: All values for this feature are identical (min === max). The data for this feature will be set to 0, effectively removing it from the model. Please check your input data.");
```

What do you think?

ziyuan-linn: Sounds good! I will work on this.

shiffman: Great! let me know when it is ready to merge! Maybe we can do a release after this one?

ziyuan-linn: Ready to merge!

shiffman:
Yay, great work @ziyuan-linn!

PR #213, opened 2024-10-10
fixed the video argument problem in the BodySegmentation constructor

### LINKED ISSUE #212: BodySegmentation (BodyPix) breaks if you rename the video variable
Hello everyone! 

On the BodySegmentation reference page under the [Quick Start](https://docs.ml5js.org/#/reference/body-segmentation?id=quick-start) chapter there is a code example https://editor.p5js.org/ml5/sketches/ruoyal-RC

There is a variable in this example called video, which stores a webcam feed. If you rename this variable to literally anything else (video2 for instance), the code stops working, throws the "ReferenceError: video is not defined" at constructor (webpack:https://ml5/src/BodySegmentation/index.js:34:17) 

It looks like the problem is here:

Screenshot 2024-10-10 at 22 33 22

The constructor does not expect any video to be passed in, so it relies on a global variable called video in the sketch

agrshch:
Fixes #212 

Although I'm not entirely sure if this.video property is actually necessary since the detection methods ask for the image/video as an argument

UPD. It turned out to be redundant, so I removed it.

### DIFF
--- a/src/BodySegmentation/index.js
+++ b/src/BodySegmentation/index.js
@@ -20,11 +20,12 @@ import { handleModelName } from "../utils/handleOptions";
 class BodySegmentation {
   /**
    * Create BodyPix.
+   * @param {String} [modelName] - A name of a model to use.
    * @param {HTMLVideoElement} [video] - An HTMLVideoElement.
    * @param {object} [options] - An object with options.
    * @param {function} [callback] - A callback to be called when the model is ready.
    */
-  constructor(modelName, options, callback) {
+  constructor(modelName, video, options, callback) {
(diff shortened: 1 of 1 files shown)


ziyuan-linn:
@shiffman I think we can have this change tag along for tomorrow's release. Thank you again @agrshch!

shiffman:
@all-contributors add @agrshch for bug and code

agrshch:
@shiffman @ziyuan-linn thank **you** so much for all your work!

shiffman:
@all-contributors add @agrshch for bug and code

Issue #210, opened 2024-10-08
BlazePose 3D keypoints


shiffman:
I'm just opening an issue here to link to and track the discussion at https://github.com/ml5js/ml5-website-v02-docsify/issues/181. It started as a documentation discussion but has moved towards ideas for changing the model output for the `z` values of the `BlazePose` model. See latest proposal: https://github.com/ml5js/ml5-website-v02-docsify/issues/181#issuecomment-2398016531. I think we can do these as "non-breaking" changes. I am also going to work on a new example showing how to use the `z` data.

ziyuan-linn:
I think we have three options:

1. Remove the z values for the named keypoints:
```js
nose: {
  x: 332.6024622758805,
  y: 265.78330263473146,
  confidence: 0.9993924452777454,
}
```
2. Nest the 3d values in another object:
```js
nose: {
  x: 332.6024622758805,
  y: 265.78330263473146,
  confidence: 0.9993924452777454,
  keypoint3D: {
    x: 0.05988978072436527,
    y: -0.5489126977664187,
    z: -0.26418375968933105,
    confidence:  0.9898926452717258
  }
}
```
3. prepend the 3d values with `3d`:
```js
nose: {
  x: 332.6024622758805,
  y: 265.78330263473146,
  confidence: 0.9993924452777454,
  3dX: 0.05988978072436527,
  3dY: -0.5489126977664187,
  3dZ: -0.26418375968933105,
  3dConfidence:  0.9898926452717258
}
```

I agree this can be a non-breaking change since removing the `z` can be viewed as a bug fix. Perhaps it should be a minor change instead of a patch?

shiffman:
Briefly re-opening this issue. I realize we had a discussion about this in a meeting but the decision was not documented here. We are going with:

1. removing the `z` property that is included with the pixel `xy` coordinate as the range is confusing and not well documented.
2. Adding a `keyPoint3D` property to each named keypoint.

I've added one last fix in #218 to address removing the `z` property.

shiffman:
This can now be closed with #218! I'll be covering this in an upcoming video tutorial about BlazePose

PR #215, opened 2024-10-11
BlazePose named 3d keypoints

ziyuan-linn:
This PR removes the z values from the named keypoints and adds a keypoint3D object to the named keypoints for BlazePose mode of BodyPose.

See https://github.com/ml5js/ml5-next-gen/issues/210 and https://github.com/ml5js/ml5-website-v02-docsify/issues/181 for additional detail and reasoning.


shiffman:
Merging, yay!

PR #218, opened 2024-10-12
Delete the unused BlazePose `z` property

shiffman:
This is a quick fix to remove the `z` property as discussed in https://github.com/ml5js/ml5-website-v02-docsify/issues/181#issuecomment-2408627029 and #210


shiffman:
Apologies that there seems to be some extraneous merge commits as part of this small, quick change. I'm not sure why this happened, but I think it won't be an issue once I "squash and merge."

ziyuan-linn (approved):

Issue #211, opened 2024-10-10
FaceMesh Triangles


shiffman:
Following up on our meeting from this morning, we discussed adding a constant with a pre-defined list of triangles to the `ml5.FaceMesh` class. I realized afterwards this is similar to the `getSkeleton()` function in `ml5.BodyPose`. One idea would then be:

```js
let faceMesh = ml5.faceMesh();
let triangles = faceMesh.getTriangles();
```

or 

```js
let triangles = faceMesh.getTriangulation();
```

But since there is only one set of triangles (rather than different ones based on `MoveNet` vs. `BlazePose`, it doesn't necessarily require a function. It could be a static constant:

```js
let triangles = ml5.faceMesh.TRIANGLES
```

or an instance constant (?)

```js
let faceMesh = ml5.faceMesh();
let triangles = faceMesh.TRIANGLES;
```

I think the static way would be more conventional JS?

Here is my example that uses the triangles which is based off of @jackbdu's teaching. Jack, just curious, where did you find this list?

https://editor.p5js.org/ima_ml/sketches/hyxD1BVVn

![image](https://github.com/user-attachments/assets/a8fa1be2-24fa-4e1c-a7e7-58eb9f829261)

We could also consider reformatting the triangles like so:

```js
const TRIANGLES = [
  { a: 127, b: 34, c: 139 },
  { a: 11, b: 0, c: 37 },
  { a: 232, b: 231, c: 120},
  ...
}
```

The current format is:

```js
const TRIANGULATION = [
    127, 34, 139, 11, 0, 37, 232, 231, 120,...
```

Also, `TRIANGLES` or `TRIANGULATION`? I think I prefer `TRIANGLES`?

@MOQN and @gohai would love to hear your thoughts along with everyone else!

gohai:
I had also very much forgotten about the precedent in `ml5.BodyPose`..! 😅

No strong feelings one way or another, but two thoughts came to mind:

- `MESH` also came to mind as a possible name
- Since `ml5.faceMesh` is a function returning a new instance, I am not sure if there is a way to also expose a constant like `ml5.faceMesh.TRIANGLES` through it?

shiffman:
[...]

PR #219, opened 2024-10-14
Added `bodySegmentation.getPartsId()`

alanvww:
@shiffman @ziyuan-linn 
As we discussed in https://github.com/ml5js/ml5-website-v02-docsify/issues/88 & https://github.com/ml5js/ml5-next-gen/pull/139 , this PR is to wrap the palette in a public method called `getPartsId()` within the `bodySegmentation`. This method will return an object containing all the constants as key-value pairs, similar to the `getSkeleton()` method in BodyPose.
CleanShot 2024-10-14 at 15 06 00@2x
As we are [removing the `properties` section from the model documentation page](https://github.com/ml5js/ml5-website-v02-docsify/pull/188), this public method ensures that the data remains accessible to all users and aligns with the consistency of other models.

### DIFF
--- a/examples/bodySegmentation-select-body-parts/sketch.js
+++ b/examples/bodySegmentation-select-body-parts/sketch.js
@@ -32,11 +32,12 @@ function draw() {
   background(255);
   image(video, 0, 0);
   if (segmentation) {
+    let parts = bodySegmentation.getPartsId()
     let gridSize = 10;
     for (let x = 0; x < video.width; x += gridSize) {
       for (let y = 0; y < video.height; y += gridSize) {
         if (
-          segmentation.data[y * video.width + x] == bodySegmentation.TORSO_FRONT
+          segmentation.data[y * video.width + x] == parts.TORSO_FRONT
         ) {
           fill(255, 0, 0);
           noStroke();
--- a/examples/bodySegmentation-select-body-parts/sketch.js
+++ b/examples/bodySegmentation-select-body-parts/sketch.js
@@ -32,13 +32,11 @@ function draw() {
   background(255);
   image(video, 0, 0);
   if (segmentation) {
-    let parts = bodySegmentation.getPartsId()
+    let parts = bodySegmentation.getPartsId();
     let gridSize = 10;
     for (let x = 0; x < video.width; x += gridSize) {
       for (let y = 0; y < video.height; y += gridSize) {
-        if (
-          segmentation.data[y * video.width + x] == parts.TORSO_FRONT
-        ) {
+        if (segmentation.data[y * video.width + x] == parts.TORSO_FRONT) {
           fill(255, 0, 0);
           noStroke();
           circle(x, y, gridSize);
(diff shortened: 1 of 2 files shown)


shiffman:
Fantastic! This will go along well with #211 as well! I hope to add that soon as I will be recording a `faceMesh` video tutorial next.

shiffman:
@gohai since this continues your work from #139 I thought I'd check with you to review before merging. Thank you!

gohai:
This looks good to me @alanvww.

One stylistic nitpick: in the example, one line is missing the trailing semicolon

In #139, we added two constants to MediaPipeSelfieSegmentation as well. I am thinking it's probably best to remove them with the change now. https://github.com/ml5js/ml5-next-gen/blob/aa05f67b9302658394e5a67b2d5aab4586d3434a/src/BodySegmentation/index.js#L189

alanvww:
> This looks good to me @alanvww.
> 
> One stylistic nitpick: in the example, one line is missing the trailing semicolon
> 
> In #139, we added two constants to MediaPipeSelfieSegmentation as well. I am thinking it's probably best to remove them with the change now.
> 
> https://github.com/ml5js/ml5-next-gen/blob/aa05f67b9302658394e5a67b2d5aab4586d3434a/src/BodySegmentation/index.js#L189

Thanks, @gohai! I’ll commit the changes soon. Regarding the constants in `SelfieSegmentation`, I double-check that there aren’t any current examples using them. However, even if we remove them, it might be a good idea to add more documentation about those properties? @shiffman @MOQN

examples/bodySegmentation-select-body-parts/sketch.js
```diff
@@ -32,11 +32,12 @@ function draw() {
   background(255);
   image(video, 0, 0);
   if (segmentation) {
+    let parts = bodySegmentation.getPartsId()
```
shiffman, line 35:
```suggestion
    let parts = bodySegmentation.getPartsId();
```

src/BodySegmentation/index.js
```diff
@@ -131,10 +131,14 @@ class BodySegmentation {
...
-        this[key] = BODYPIX_PALETTE[key].id;
-      }
+      // Add function to return parts' key and id
+      this.getPartsId = () => {
+        let parts = {};
+        for (let key in BODYPIX_PALETTE) {
+         parts[key] = BODYPIX_PALETTE[key].id;
+       }
+       return parts;
+     }
```
shiffman, line 141: Tiny thing of course but looks like the indents may be off here :)

shiffman (commented):
Looks great to me! I agree we can remove the selfie segmentation constants. Normally we might "deprecate" the property, but I did a search and they appear nowhere in the examples or documentation so I think it's good to remove.

alanvww:
@gohai @shiffman Thanks for the reviews! I’ve just made the changes to the code format, removing the selfie segmentation constants.

shiffman:
Fantastic work, thank you! Merging! I may have some more thoughts or questions when I get to recording a video tutorial about this, stay tuned!

PR #222, opened 2024-10-19
add `FaceMesh.getTriangles`

ziyuan-linn:
This PR adds the `FaceMesh.getTriangles` function. It provides a nested array of keypoint indices. The indices are grouped in threes, indicating which three keypoints form a triangle. 

Changes:
- Implement `getTriangles` function
- Add triangle mesh example

See #211 for details.


shiffman:
Merging and we can continue the discussion in #211!

Issue #224, opened 2024-10-23
HandPose Connections


shiffman:
Picking up on #231, I would like to propose adding `getConnections()` to `ml5.handPose()`. According to the official diagram, these are the connections.  (I made this manually and the order is somewhat arbitrary. Is there an official standard?)

```js
let connections = [
  [0, 1],
  [1, 2],
  [2, 3],
  [3, 4],
  [0, 5],
  [5, 6],
  [6, 7],
  [7, 8],
  [9, 10],
  [10, 11],
  [11, 12],
  [13, 14],
  [14, 15],
  [15, 16],
  [0, 17],
  [17, 18],
  [18, 19],
  [19, 20],
  [5, 9],
  [9, 13],
  [13, 17],
];
```

<img width="551" alt="Screenshot 2024-10-23 at 3 35 51 PM" src="https://github.com/user-attachments/assets/b3fd92b0-adda-4207-b1ca-836ccb452c8f">

ziyuan-linn:
This looks great! I also think it is a good idea to add connections for `handPose`.  

Looks like the official standards are defined in [constants.ts](https://github.com/tensorflow/tfjs-models/blob/6f6a19848415a4ba564332232a9d869d4e41722b/hand-pose-detection/src/constants.ts#L50). I think we can import it with `import { MEDIAPIPE_CONNECTED_KEYPOINTS_PAIRS } from "@tensorflow-models/hand-pose-detection/dist/constants";`.

Going back to the previous discussion, perhaps we can also add `bodyPose.getConnections` as an alias for `bodyPose.getSkeleton` and eventually deprecate the latter?

ziyuan-linn:
@shiffman I just made a PR to add the `getConnections` function.

I noticed that there is a small difference between the official connections array and the diagram. This is from the connections array: 
![hand skeletal connections drawn from official connections array](https://github.com/user-attachments/assets/0aa015b1-5bf7-4ccf-958f-02a714d93e0f)

Which of the two connection schemes do you think we should use?

shiffman:
Oh interesting! I think it probably makes sense to follow the official ones! It's just strange that the official connections don't match the diagram? It's not so many keypoints that someone couldn't easily build their own list if they prefer a different visual. Maybe @MOQN has a preference?

![image](https://github.com/user-attachments/assets/1821b902-fbd5-4666-8ccf-86f4a6e99da3)

#227

MOQN:
[...]

PR #226, opened 2024-10-29
HandPose connections

ziyuan-linn:
This PR implements the `getConnections` function for handPose and adds a skeletal connection example. See #224 for details. 

One small question that remains is how shall we name the examples now that we are switching to `getConnections` instead of `getSkeleton`. The current examples are named `modelName-skeleton`. Perhaps we could rename them `modelName-skeletal-connections`?


shiffman:
Merging, yay!

PR #242, opened 2025-02-06
Add support for OffscreenCanvas

quinton-ashley:
Tensorflow will accept "canvas like" objects, including OffscreenCanvas.

https://github.com/tensorflow/tfjs/blob/2644bd0d6cea677f80e44ed4a44bea5e04aabeb3/tfjs-core/src/ops/browser.ts#L98

I tested if this with the ml5 image classifier example and it does work!

### DIFF
--- a/src/utils/handleArguments.js
+++ b/src/utils/handleArguments.js
@@ -7,7 +7,7 @@ import * as tf from "@tensorflow/tfjs";
 /**
  * Standard input accepted by most TensorFlow models.
  * @typedef InputImage
- * @type {ImageData | HTMLImageElement | HTMLCanvasElement | HTMLVideoElement | tf.Tensor3D}
+ * @type {ImageData | HTMLImageElement | HTMLCanvasElement | OffscreenCanvas | HTMLVideoElement | tf.Tensor3D}
  */
 
 /**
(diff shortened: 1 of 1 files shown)


shiffman:
Thank you @quinton-ashley! This makes sense to include I think! Checking with @gohai and @ziyuan-linn before I merge in case they have any thoughts or comments.

gohai:
This looks good to me on its own. I wasn't aware of OffscreenCanvas even, but it's been widely supported since 2023, so iOS 16.4 and higher should have this class defined.

I saw that [darknet](https://github.com/ml5js/ml5-next-gen/blob/main/src/ImageClassifier/darknet.js#L31) and [doodlenet](https://github.com/ml5js/ml5-next-gen/blob/main/src/ImageClassifier/doodlenet.js#L20) currently check against the input type explicitly. Perhaps this would be a good opportunity to have them use `isImageElement` from handleArguments.js?

@quinton-ashley Have you tried some of the other ml5 examples, e.g. `faceMesh-*`? Some of the code we build upon doesn't seem to take OffscreenCanvas into consideration, so I'd be curious how well this works.

ziyuan-linn:
I also think this change makes sense!

PR #248, opened 2025-05-01
Adding tensorflow.js depth-estimation

alanvww:
Hello! This PR is adding the [depth estimation](https://github.com/tensorflow/tfjs-models/tree/master/depth-estimation/src/ar_portrait_depth) functionality from tensorflow.js to ml5.

I primarily refer to this [example](https://storage.googleapis.com/tfjs-models/demos/3dphoto/index.html) for its performance and results.

Testing sketches:
[depthEstimation-video](https://editor.p5js.org/alanvww/sketches/3ag-ff9vg)
[depthEstimation-single-image](https://editor.p5js.org/alanvww/sketches/EZLyLb7Ez)

image

- [x] Set `grayscale` colormap as default
- [ ] Remove bodySegmentation? 
- [ ] Backend option to use transformers.js(!)


examples/depthEstimation-single-image/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>ml5.js | Depth Estimation Single Image</title>
```
gohai: Nitpick: the existing examples all have their title in such a format fwiw: `ml5.js faceMesh Bounding Box Example`

examples/depthEstimation-single-image/index.html
```diff
@@ -0,0 +1,13 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.9.4/p5.min.js"></script>
+    <script src="../../dist/ml5.js"></script>
+  </head>
+
+  <body>
+    <script src="sketch.js"></script>
+  </body>
+</html>
```
gohai, line 15: Since git is operating on entire lines, it expects the last line of code to still be followed by a newline character - that's what this warning is about. (Just pressing enter one more time here and in some other files should address this.)

examples/depthEstimation-single-image/sketch.js
```diff
@@ -0,0 +1,137 @@
+/*
+ * 👋 Hello! This is an ml5.js example made and shared with ❤️.
+ * Learn more about the ml5.js project: https://ml5js.org/
+ * ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
+ *
+ * This example demonstrates depth estimation on a single image using ml5.depthEstimation.
+ */
+
+// Global variables
```
gohai, line 9: Maybe superfluous

examples/depthEstimation-single-image/sketch.js
```diff
@@ -0,0 +1,137 @@
...
+let depthResult;    // The object holding the depth estimation results
+let status = "Loading model and image..."; // Status message
+
+// Image dimensions (will be updated based on loaded image)
+let imgWidth = 640; // Default width
+let imgHeight = 480; // Default height
+
+// Model and runtime options
+let options = {
+    colormap: "COLOR", // Colormap for visualization ('COLOR', 'GRAYSCALE')
```
gohai: Seconding Dan's suggestion to use GREYSCALE as the default, as it is conceptually simpler

examples/depthEstimation-single-image/sketch.js
```diff
@@ -0,0 +1,137 @@
...
+    minDepth: 0,       // Minimum depth value (0-1) for normalization
+    maxDepth: 1,       // Maximum depth value (0-1) for normalization
+};
+
+// We don't use preload anymore, setup will handle async loading
+// function preload() {}
```
gohai: Is this a choice or a requirement for the depth estimation feature, I wonder?

For an immediate release, it might be more understandable if the example worked the same way the other examples did... if we're thinking that this will get released together with the p5.js 2.0 work, then this is point is probably largely moot

examples/depthEstimation-single-image/sketch.js
alanvww: > Maybe superfluous

😂The PR might have a lot of minor issues like this. For context, the first commit is a squashed version of work I did about 2 months ago, while the second commit contains changes I made mainly in the last 10 days with Copilot's help. I'll definitely take some time to clean things up and address issues. Your thorough review is greatly appreciated!🙏

examples/depthEstimation-single-image/sketch.js
```diff
@@ -0,0 +1,137 @@
...
+
+async function setup() {
+    // Create a canvas twice the width of the image to show image and depth map side-by-side
+    // Initial size is default, will be resized later
+    createCanvas(imgWidth * 2, imgHeight);
+    console.log("Setup started.");
```
gohai: The existing examples tend to keep console messages to a minimum, since often times with the large font sizes people tend to use in educational (projected) settings, the amount of lines visible at the same time is quite limited - so keeping it terse has the benefit of making it possible for people to see the more of the overall structure of the example at once.

And on a super nitpick note: we tend to not have tailing dots in console messages

examples/depthEstimation-single-image/sketch.js
```diff
@@ -0,0 +1,137 @@
...
+
+    // Initialize the depth estimation model
+    console.log("Loading depth estimation model...");
+    depthEstimator = ml5.depthEstimation(options);
+
+    // Load the sample image using a promise wrapper for await
+    console.log("Loading image...");
+    try {
```
gohai: None of the other examples currently use exception handling, and this might be a technique that less experienced users aren't yet familiar with. If this is for the promise, then `preload` could be simpler to understand?

examples/depthEstimation-single-image/sketch.js
```diff
@@ -0,0 +1,137 @@
...
+        // --- IMPORTANT: Replace with the path to YOUR image ---
+        img = await loadImagePromise('face.png'); // Use the user's image
+        // img = await loadImagePromise('../../3d_photo/images/im1.jpg'); // Original image path
```
gohai: Since the `face.png` is a file in the current folder, spelling those steps out in such detail might be unnecessary

examples/depthEstimation-single-image/sketch.js
alanvww: This was for quick testing of something that did not work with the depth estimation library before. But now it is fixed! And this commit is outdated, please check the newer commit as well! The version of the first commit is an even bigger mess lol.

examples/depthEstimation-single-image/sketch.js
```diff
@@ -0,0 +1,137 @@
...
+
+        // Update dimensions based on the loaded image
+        imgWidth = img.width;
+        imgHeight = img.height;
+        resizeCanvas(imgWidth * 2, imgHeight); // Resize canvas now
+
+        // Wait for the model to be ready
+        status = "Waiting for model to load...";
+        console.log(status);
+        await depthEstimator.ready; // Wait for the promise returned by the constructor
```
gohai: `.ready` or not - probably something to think about in the context of the p5 2.0 work 👍 thanks for prototyping all of this, Alan!

examples/depthEstimation-single-image/sketch.js
```diff
@@ -0,0 +1,137 @@
...
+            if (img) {
+                resolve(img);
+            } else {
+                reject(`Failed to load image at path: ${path}`);
+            }
+        }, err => {
+            reject(`Error loading image: ${err}`);
+        });
+    });
+}
```
gohai, line 26: Will this helper function be obsolete with p5 2.0, I wonder?
I'd normally try to do without generic helper functions, that aren't intimately tied to the aspect or functionality this example is dealing with - just because users who try to read and understand might have more opportunities to get unnecessarily confused? (e.g. "What is a Promise? What is this resolve and reject function?")

examples/depthEstimation-single-image/sketch.js
```diff
@@ -0,0 +1,137 @@
...
+        textAlign(CENTER, CENTER);
+        textSize(16);
+        text(status, imgWidth / 2, imgHeight / 2);
+        textAlign(LEFT, BASELINE); // Reset text alignment
+    }
+
+    // Check if depth estimation results are available
+    if (depthResult) {
+        // Draw the colormapped depth visualization on the right half of the canvas
+        image(depthResult.visualizationImage, imgWidth, 0, imgWidth, imgHeight);
```
gohai: `visualizationImage` seems a bit long - is this the naming that the tensorflow.js model uses?

examples/depthEstimation-single-image/sketch.js
```diff
@@ -0,0 +1,137 @@
...
+        textSize(16);
+        text(status, imgWidth + imgWidth / 2, imgHeight / 2);
+        textAlign(LEFT, BASELINE); // Reset text alignment
+    }
+}
+
+// Callback function that receives the depth estimation results
+function gotResults(result) {
+    // Store the result in the global variable (already done by await in estimateDepth)
+    // depthResult = result;
```
gohai: With `await`, it might be clearer to forgo the "callback" function entirely, and do it all in the caller?

examples/depthEstimation-single-image/sketch.js
```diff
@@ -0,0 +1,137 @@
+/*
```
gohai, line 1: This file uses four spaces for indentation, while elsewhere in the library, we tend to use two. If you run `yarn format`, this should take care of prettifying this for you automatically.

alanvww: > This file uses four spaces for indentation, while elsewhere in the library, we tend to use two. If you run `yarn format`, this should take care of prettifying this for you automatically.

Thanks! That is good to know.

examples/depthEstimation-video/sketch.js
```diff
@@ -0,0 +1,110 @@
...
+    // Create the video capture element
+    video = createCapture(VIDEO);
+    video.size(videoWidth, videoHeight); // Set video size
+    video.hide(); // Hide the default HTML video element
+
+// Create buttons for changing colormaps
+    const colormaps = ['COLOR', 'GRAYSCALE']; // Updated available colormaps
+    let buttonY = videoHeight + 15; // Position buttons below the canvas
+    let buttonX = 10;
+    colormaps.forEach(mapName => {
```
gohai: None of the existing examples presently use `forEach` or arrow functions, which might be constructs less experienced users aren't familiar with. Could be good to do this in a slightly more pedestrian (simpler) way instead.

examples/depthEstimation-single-image/sketch.js
alanvww: > `visualizationImage` seems a bit long - is this the naming that the tensorflow.js model uses?

If I remember correctly, it's the naming in one of their examples. But it's totally changeable!

examples/depthEstimation-video/sketch.js
```diff
@@ -0,0 +1,110 @@
...
+// Model and runtime options
+let options = {
+    colormap: "COLOR", // Colormap for visualization ('COLOR', 'GRAYSCALE')
+    minDepth: 0,       // Minimum depth value (0-1) for normalization
+    maxDepth: 1,       // Maximum depth value (0-1) for normalization
+    // flipHorizontal: false // Set to true if using a webcam that needs flipping
+};
+
+// Video dimensions
+const videoWidth = 640;
```
gohai, line 14: There are some examples that use `const`, but we currently are using `let` in many places where `const` would be appropriate - fyi.

examples/depthEstimation-video/sketch.js
```diff
@@ -0,0 +1,110 @@
...
+    video.hide(); // Hide the default HTML video element
+
+// Create buttons for changing colormaps
+    const colormaps = ['COLOR', 'GRAYSCALE']; // Updated available colormaps
+    let buttonY = videoHeight + 15; // Position buttons below the canvas
+    let buttonX = 10;
+    colormaps.forEach(mapName => {
+        let button = createButton(mapName);
+        button.position(buttonX, buttonY);
+        button.mousePressed(() => changeColormap(mapName));
```
gohai: (additional arrow function)

examples/depthEstimation-video/sketch.js
alanvww: Yes! I totally agree. It was there for swapping different colormaps, but now it only has color and grayscale and change by tweaking the options.

examples/depthEstimation-video/sketch.js
```diff
@@ -0,0 +1,110 @@
...
+// Optional: Log the full result object when the mouse is pressed
+function mousePressed() {
+    if (depthResult) {
+        console.log("Current Depth Result:", depthResult);
+    }
+}
+// Function to change the colormap and re-initialize the model
+// Function to change the colormap dynamically
+function changeColormap(newColormap) {
+    console.log(`Changing colormap to: ${newColormap}`);
```
gohai: Currently none of the existing examples use template literals.

(I am not familiar with p5 2.0 yet, so I don't know if this is extensively making use of them. In case it isn't, I would stick to concatenating strings with `+`, since its syntactically simpler.)

examples/depthEstimation-video/sketch.js
```diff
@@ -0,0 +1,110 @@
...
+    if (depthEstimator && depthEstimator.runtimeConfig) {
+        // Directly update the runtime configuration of the existing estimator
+        // Ensure the value is uppercase as the library expects
+        depthEstimator.runtimeConfig.colormap = newColormap.toUpperCase();
```
gohai: Could be better to have this (tolerant) logic in the library, rather than the example(s)

src/DepthEstimation/index.js
```diff
@@ -0,0 +1,526 @@
...
+ * @property {number} maxDepth - The maximum depth value used for normalization during processing.
+ * @property {function(number, number): number | null} getDepthAt - A function to get the raw depth value at a specific (x, y) coordinate. Returns `null` if coordinates are out of bounds. Coordinates are relative to the depth map dimensions.
+ */
+
+/**
+ * @class
+ * DepthEstimation class using the TensorFlow.js ARPortraitDepth model.
+ * Estimates depth from a single image.
+ *
+ * @example
```
gohai: Since the examples in the same repo will get more testing than those Javadoc snippets, I would leave those out, as they very easily get outdate (e.g. this mentions a colormap "viridis")

examples/depthEstimation-single-image/sketch.js
gohai: @shiffman and @MOQN are very thoughtful and gifted when it comes to naming!

src/DepthEstimation/index.js
nasif-co: Resolved in https://github.com/alanvww/ml5-next-gen/commit/41d354114bbd46fa27e973990860fa3f2d9dcbe2 :)

examples/depthEstimation-single-image/sketch.js
nasif-co: Resolved in https://github.com/ml5js/ml5-next-gen/pull/248/commits/fd74d0b18f2fb60f7754112f97263347491a0712

src/DepthEstimation/index.js
```diff
@@ -0,0 +1,790 @@
...
+        result.minDepth,
+        result.maxDepth,
+        currentRuntimeConfig.colormap
+      );
+
+      // --- Apply Black Background using Segmentation Mask (if enabled) ---
+      if (currentRuntimeConfig.applySegmentationMask && this.segmenter) {
+        try {
+          // Run segmentation on the *original* source element to get the mask
+          const segmentation = await this.segmenter.segmentPeople(
```
nasif-co, line 435: Segmentation of the source image was done before on line 267, I'll work on reusing that result here. Also maybe worth it to use a [dilation filter](https://gist.github.com/nasif-co/76a8b894d71e06be6976f7c8a7db0262#dealing-with-edge-noise) here to further clean up the data before returning it to the user.

examples/depthEstimation-single-image/sketch.js
nasif-co: Previously resolved in https://github.com/ml5js/ml5-next-gen/commit/41d354114bbd46fa27e973990860fa3f2d9dcbe2 :)

examples/depthEstimation-single-image/sketch.js
nasif-co: Replaced with preload in https://github.com/ml5js/ml5-next-gen/commit/41d354114bbd46fa27e973990860fa3f2d9dcbe2, still important to think of this regarding p5 2.0 compatible examples

examples/depthEstimation-video/sketch.js
nasif-co: Changed in https://github.com/ml5js/ml5-next-gen/commit/41d354114bbd46fa27e973990860fa3f2d9dcbe2

examples/depthEstimation-single-image/sketch.js
nasif-co: Addressed in https://github.com/ml5js/ml5-next-gen/pull/248/commits/ad8c12e575f416adfe65aaf6ee5af099d63275aa :)

examples/depthEstimation-video/sketch.js
shiffman: I think we should follow p5.js style which (unless it's changed?) is to use `let` everywhere. That said, I'm open minded if you think otherwise!

examples/depthEstimation-video/sketch.js
shiffman: Also, there are elements of this sketch that are more "demo" than "example", it is likely be useful to have an additional example like this but I would suggest starting with a very basic "hello world depth estimation" that uses the defaults (minimal custom `options`), only shows the depth image in the canvas (not both), and doesn't have any buttons or interface. And of course @nasif-co I think there's a lot of room for some basic 3D rendering examples along the lines of your projects from spring 25!

nasif-co:
> Is this an issue only if you resize the video _during_ a sketch, or if you just call `video.size(w, h)` once in `setup()` does it break the depth map?

It happens by calling  `video.size(w, h)` once in `setup()`, you can see a reproduction of the issue in [this sketch](https://editor.p5js.org/nasif-co/sketches/TlOCC-aD1).

> * Am I right that darker pixels are closer to the camera and brighter are further? This is the opposite of my expectation since I think the transformers.js models work the other way. Nothing to change here, just noting it as something to mention in documentation!

Yes you are right, it confused me a bit at first. Now that I'm looking into it, changing it to be the other way around seems to be a small change that could be useful in keeping consistency with transformers.js, looking ahead at integrating it. I'll commit that small change.

> * The "hello world" examples are perfect, exactly as I imagined! Now after seeing them I'm wondering if we might consider including a single example that incorporates something with 3D or uses the pixel data in some way? Perhaps a nested loop through every N pixels and draw a `box()` for each value with a z-position mapped to the pixel value?

Yes! I was looking at including some of those next. I had this [sketch](https://editor.p5js.org/nasif-co/sketches/h3WR7pLn2) I made a few months ago for class using transformers.js which sounds like what you're describing. I'll port that one to ml5. Do you think we should just do the one? I was thinking of also adding one that builds a 3D mesh using the depth map, but I don't know if that becomes more of a tutorial territory than example.

Also was planning on adding an example that showcases how to "detect" distance, so that different interactions can occur depending on how close/far a subject is. Something like a chain of if/else statements, each with a different interaction. I suspect this would be a common use case of the depth estimation model.

> * I think the example that uses the `mask()` might be more effective if there is an image or maybe something simple drawn behind the silhouette, it's not so clear what is happening with only `clear()`!

I agree, I'll add something simple in the back, maybe a background color shifting in hue or just an image.

Thanks for the detailed review! :)

nasif-co:
Interestingly, the bug also affects the body segmentation module, when using `SelfieSegmentation` (see a [sketch of it](https://editor.p5js.org/nasif-co/sketches/RDZcXTRlw)) but **not** when using `BodyPix`, which is strange since both use the same function to do the detection.

On the other hand, it makes some sense since the `ARPortraitDepth` model we are using here also uses `SelfieSegmentation`, so that may be the root of the issue. I feel it has something to do with the source video's [intrinsic dimensions](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/videoHeight#about_intrinsic_width_and_height) as opposed to its display dimensions, and how the `video.size()` method [only sets display width](https://github.com/processing/p5.js/blob/b9cdbd79987f333ccdeb1f4609657eed69d7131d/src/dom/dom.js#L3711-L3715). The one way I have found to change intrinsic dimensions of the webcam video element is to request them with `getUserMedia` when creating the capture, which is out of the question.

Going forward, I think the best solution is rendering the video pixels to a separate canvas/p5.Graphics in ml5 and passing that as the `detectMedia`. Would love to hear some thoughts on this!

examples/depthEstimation-video/sketch.js
nasif-co: Since the video example also uses the webcam, I modified it in https://github.com/ml5js/ml5-next-gen/pull/248/commits/5f0e3fb177962b557507baaf6af7d51f6ecea735 to instead showcase how to run depthEstimation with a video file. Other examples have been added in previous commits.

nasif-co:
Updated the examples to add the p5 2.0 version. Interestingly, the point cloud example had a performance drop, and the mesh example had a great performance boost. Looking into it in https://github.com/processing/p5.js/issues/6438, it must be related to the mesh example making use of `p5.Geometry` while the point cloud just uses 3D primitives. May be a good idea to modify the point cloud one in the future to use `p5.Geometry` instead.

PR #250, opened 2025-05-05
Turn "const" into "let" in example code

gohai:
The p5 reference does not currently introduce "const", and the containing example codes don't make use of it either (as far as I could tell). This is converting all our example code to "let".

There are only a few places where "const" is used, and its application there isn't very consistent either [1], so I think removing those for now I fine.

[1] e.g. in neuralNetwork-load-model, `classifierOptions` is `let` but `modelDetails` is `const`


shiffman:
Thank you @gohai!!

Issue #249, opened 2025-05-03
index.html files in examples-dir don't work: they try to load ../../dist/ml5.js which doesn't exist


rajeevraizada:
### Issue Type

Runtime Error

### ml5.js Version

ml5-next-gen

### Development Environment

VS Code live server

### Browser & Operating System

Chrome 135 on Mac OS 14.7.4

### Issue Description

Need to replace this line in the index.html files
<script src="../../dist/ml5.js"></script>
with this line:
<script src="https://unpkg.com/ml5@1/dist/ml5.min.js"></script>
or, altenatively, put a dist directory into the ml5-next-gen github repository, so that the ../../dist/ml5.js line points to a file that is actually there.

### Minimal Reproducible Example

I downloaded the ml5-next-gen repository from github, and tried running the examples in the examples directory. They all got stuck on "Loading..."

It turns out that the index.html files in the examples all try to load ml5 locally, using this line:
<script src="../../dist/ml5.js"></script>

However, there is no dist directory in the root of the github version of ml5-next-gen (retrieved May 3, 2025).

When I replaced that local ml5 sourcing line (which points to a non-existent location) with loading from the actual online server:
<script src="https://unpkg.com/ml5@1/dist/ml5.min.js"></script>
then everything worked fine.

This looks like it should be an easy fix, fortunately. 

Raj

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

gohai:
Hello @rajeevraizada - the `dist/ml5.js` gets automatically generated by building the library. You can find the steps to do so in the [CONTRIBUTING.md](https://github.com/ml5js/ml5-next-gen/blob/main/CONTRIBUTING.md#building-the-library). We should probably clarify the Readme to this end. Hope this helps!

rajeevraizada:
Aha, thanks! I agree that a prominent note in the Readme would be useful here. I was not expecting the downloaded repository to be missing the actual ml5.js file.

That is probably due to my ignorance, but I wonder if it might be an expectation shared by a lot of other people too: I didn't think that JS libraries needed to be built or compiled at all. I thought that was only required for languages that need to get turned into some sort of device-dependent binary form, like C, or Python packages if they contain compiled executables. I thought that with JS it is uncompiled scripts, all the way down?

Would there be any downside to simply including the actual ml5.js script as a pre-existing file already in this repository? Unless different platforms require different versions of it, then that would seem like a useful addition.

[...]

PR #251, opened 2025-05-08
Add a Readme file to examples, with link to sketches on web editor

gohai:
This makes it easier to go from the repository to running the example sketches, without setting up a local development environment. (See #249 for discussion) In the future, it would be a nice addition to the uploadExamples.js script, if it'd programmatically update the Markdown file.


shiffman:
This is great! It occurs to me that it might be nice to have a single page on the website with all of the examples! I built something like this for the nature of code book which I find incredibly useful: https://natureofcode.com/examples/ cc @MOQN

PR #252, opened 2025-05-08
WIP Timeseries base

mop9047:
Adds a TimeSeries Feature so sequences of data can be used in ML5

This version extends from the Neural Network class with some overrides to allow sequences of data. It also includes some helper functions to simplify loading timeseries data.

Key Features:

- `addData` in the format of Array of Objects [{x: ,y: },{x: ,y: }] or Array of Arrays [[,],[,]]
- `addDefaultLayers` which uses `tsLayers.js` file to define the architectures, based on `task` and `dataMode` (still needs to be discussed).
- `padCoordinates(sequence, targetLength)` uses RDP algorithm to simplify coordinate data to a specified target length
- `slidingWindow(data, featureKeys, targetKeys, batchLength = null)` and `sampleWindow(data)` uses sliding Window algorithm to create batches from sequence data and take a single sample for prediction


src/TimeSeries/index.js
```diff
@@ -0,0 +1,193 @@
...
+    this.neuralNetworkData =
+      this.options.neuralNetworkData || new TimeSeriesData();
+
+    this.init = this.init.bind(this);
+    this.ready = callCallback(this.init(), callback);
+  }
+
+  async init() {
+    // workaround for Error
+    setBackend("webgl");
```
gohai, line 36: Regular NeuralNetwork leaves this to the sketch to set, so I'd suggest we do the same. (Or, change NN in the same manner. I don't recall, but there might have been previous discussion why we didn't transparently forced the backend to "webgl" in such a way.)

src/TimeSeries/index.js
```diff
@@ -0,0 +1,193 @@
...
+      case "regression_spatial":
+        layers = tsLayers.regression_spatial;
+        return this.createNetworkLayers(layers);
+
+      case "regression_linear":
+        layers = tsLayers.regression_linear;
+        return this.createNetworkLayers(layers);
+
+      default:
+        console.log("no inputUnits or outputUnits defined");
```
gohai, line 120: console.warn(), console.error()?

src/TimeSeries/index.js
```diff
@@ -0,0 +1,193 @@
...
+        optimizer: tf.train.adam,
+        metrics: ["accuracy"],
+      };
+    } else if (this.options.task === "regression") {
+      options = {
+        loss: "meanSquaredError",
+        optimizer: tf.train.adam,
+        metrics: ["accuracy"],
+      };
+    }
```
gohai, line 146: Maybe a catch-all else that informs the user that they specified an unknown/unsupported task?

src/TimeSeries/timeSeriesData.js
```diff
@@ -0,0 +1,381 @@
...
+  }
+
+  /**
+   * convertRawToTensors
+   * converts array of {xs, ys} to tensors
+   * @param {*} dataRaw
+   *
+   * @return {{ inputs: tf.Tensor, outputs: tf.Tensor }}
+   */
+  // eslint-disable-next-line class-methods-use-this, no-unused-vars
```
gohai, line 92: Is this needed (for our prettifier)?

examples/timeSeries-hand-gestures/sketch.js
```diff
@@ -0,0 +1,192 @@
...
+  model.normalizeData();
+  let options = {
+    epochs: 100,
+  };
+  model.train(options, whileTraining, finishedTraining);
+  nameField.value("");
+}
+
+function whileTraining(epoch) {
+  console.log(epoch);
```
gohai, line 135: Easy to get confused with console.log()s that aren't your (the user's) own... :)

Best to make them as descriptive as possible (e.g. `console.log("In epoch", epoch);` and then commenting them out to, so that they're still available for future debugging.

src/TimeSeries/timeSeriesData.js
```diff
@@ -0,0 +1,381 @@
...
+      // reshape array - already ready for tensorconversion
+      const features = Object.keys(inputMeta);
+      const feature_length = features.length;
+
+      const seriesStep = dataRaw[0]["xs"].length;
+
+      const batch = normalized[features[0]].length / seriesStep;
+
+      this.meta.seriesShape = [seriesStep, feature_length];
+      console.log("series shape", this.meta.seriesShape);
```
gohai, line 223: Easy to get confused with console.log()s that aren't your (the user's) own... :)

Best to make them as descriptive as possible (e.g. console.log("In epoch", epoch); and then commenting them out to, so that they're still available for future debugging.

src/TimeSeries/timeSeriesData.js
```diff
@@ -0,0 +1,381 @@
...
+  getDTypesFromData() {
+    const meta = {
+      ...this.meta,
+      inputs: {},
+      outputs: {},
+    };
+
+    const sample = this.data.raw[0];
+
+    //consistent dTypes have already been checked at add data
```
gohai, line 32: Nitpick: space after `//`

src/TimeSeries/timeSeriesData.js
```diff
@@ -0,0 +1,381 @@
...
+    const xs = Object.keys(sample.xs[0]); //since time series data is in form of array
```
gohai, line 33: Nitpick: space after `//`

examples/timeSeries-hand-gestures/sketch.js
```diff
@@ -0,0 +1,192 @@
...
+
+function preload() {
+  // Load the handPose model
+  handPose = ml5.handPose();
+
+  // setup the timeseries neural network
+  let options = {
+    outputs: ["label"],
+    task: "classification",
+    dataMode: "spatial",
```
gohai, line 31: If my memory serves me right: I thought the plan was to drop the extra property to activate the special handling for spatial data, and instead fold it into `task` (similarly to `classification` and `imageClassification`)?

examples/timeSeries-load-model-hand-gestures/model/model.json
```diff
@@ -0,0 +1 @@
...arizer":null,"activity_regularizer":null,"kernel_constraint":null,"bias_constraint":null,"name":"dense_Dense1","trainable":true}},{"class_name":"Dense","config":{"units":2,"activation":"softmax","use_bias":true,"kernel_initializer":{"class_name":"VarianceScaling","config":{"scale":1,"mode":"fan_avg","distribution":"normal","seed":null}},"bias_initializer":{"class_name":"Zeros","config":{}},"kernel_regularizer":null,"bias_regularizer":null,"activity_regularizer":null,"kernel_constraint":null,"bias_constraint":null,"name":"dense_Dense2","trainable":true}}]},"keras_version":"tfjs-layers 4.8.0","backend":"tensor_flow.js"},"weightsManifest":[{"paths":["./hello.weights.bin"],"weights":[{"name":"conv1d_Conv1D1/kernel","shape":[3,42,8],"dtype":"float32"},{"name":"conv1d_Conv1D1/bias","shape":[8],"dtype":"float32"},{"name":"conv1d_Conv1D2/kernel","shape":[3,8,16],"dtype":"float32"},{"name":"conv1d_Conv1D2/bias","shape":[16],"dtype":"float32"},{"name":"dense_Dense1/kernel","shape":[176,16],"dtype":"float32"},{"name":"dense_Dense1/bias","shape":[16],"dtype":"float32"},{"name":"dense_Dense2/kernel","shape":[16,2],"dtype":"float32"},{"name":"dense_Dense2/bias","shape":[2],"dtype":"float32"}]}]}
```
gohai, line 1: Nitpick: adding a newline at the end of the file makes Git happy

examples/timeSeries-weather-prediction/sketch.js
```diff
@@ -0,0 +1,183 @@
+/*
+ * 👋 Hello! This is an ml5.js example made and shared with ❤️.
+ * Learn more about the ml5.js project: https://ml5js.org/
+ * ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
+ *
+ * This example demonstrates Training a Stock Price Predictor through ml5.TimeSeries.
```
gohai, line 6: (This still refers to stock prices, fyi)

examples/timeSeries-hand-gestures/sketch.js
```diff
@@ -0,0 +1,192 @@
...
+    debug: "true",
+    learningRate: 0.001,
```
gohai, line 33: A comment explaining whether/why deviating from the default was necessary here, and how came up with this value, might be instructive. (e.g. "the default learning rate of 0.01 didn't converge with this data set - lowering the ...")

examples/timeSeries-hand-gestures/sketch.js
```diff
@@ -0,0 +1,192 @@
...
+  if (hands.length > 0 && recordingFinished == false) {
+    if (sequence.length <= seqLength) {
+      // get coordinates from hands (21 points)
+      handpoints = drawPoints();
+      sequence.push(handpoints);
+
+      // once sequence reaches the seqLength, add sequence as just one X value
+    } else if (sequence.length > 0) {
+      // get the training word from the input box
+      let train_word = nameField.value();
```
gohai, line 69: Stylistic nitpick: perhaps `trainWord`, since the rest of the example uses camelCase

src/TimeSeries/index.js
```diff
@@ -0,0 +1,193 @@
...
+
+  createMetaData() {
+    // this method does not get shape for images but instead for timesteps
+    const { inputs } = this.options;
+
+    let inputShape;
+    if (typeof inputs === "number") {
+      inputShape = inputs;
+    } else if (Array.isArray(inputs) && inputs.length > 0) {
+      inputShape = inputs.length; //will be fed into the tensors later
```
gohai, line 84: Nitpick: space after `//`

examples/timeSeries-hand-gestures/sketch.js
```diff
@@ -0,0 +1,192 @@
...
+  console.log(Object.keys(trainingWords));
+  if (!(tempWord in trainingWords)) {
+    trainingWords[tempWord] = 1;
+  } else {
+    trainingWords[tempWord]++;
+  }
+
+  let counts = "";
+  let keys = Object.keys(trainingWords);
+  console.log("keys", keys);
```
shiffman, line 185: Just noting this code which includes `Object.keys()` and a `for in` loop is not typical of p5.js style and may be confusing to beginners. They may be necessary in which case I might suggest explanatory comments. Also, I would probably remove the console logs from the examples.

examples/timeSeries-weather-prediction/sketch.js
```diff
@@ -0,0 +1,183 @@
...
+function preload() {
+  json_data = loadJSON("weather_data.json");
+
+  // set the options to initialize timeSeries Neural Network
+  let options = {
+    task: "regression",
+    dataMode: "linear",
+    debug: "true",
+    learningRate: 0.01,
+    output: ["label"],
```
shiffman, line 27: I'm confused by this, it doesn't match up to me with the format of the data? Also, ml5.js will infer inputs and outputs from the data itself so maybe unecessary to include?

shiffman (commented):
Some additional comments along the lines of @gohai's review!  Also, I see this PR still includes `ml5.timeSeries()`, my memory is shaky of course, but I recall a discussion around this implementation being a part of `ml5.neuralNetwork()` with a new "time series" task? Is that correct?

gohai:
> Some additional comments along the lines of @gohai's review! Also, I see this PR still includes `ml5.timeSeries()`, my memory is shaky of course, but I recall a discussion around this implementation being a part of `ml5.neuralNetwork()` with a new "time series" task? Is that correct?

My memory is also shaky, @shiffman - but I like the idea. Since the options are being passed to the factory method, we could _keep_ the inheritance model that Mathew built, and have `ml5.neuralNetwork()` simply instantiate the specialized class whenever encounters a related `task` option. The alternative is to integrate Mathew's work even more tightly into the existing NeuralNetwork classes. Do you have a preference @mop9047?

src/TimeSeries/index.js
mop9047: done :) I chose to do console.warn instead as it's not necessarily an error and the layers are chosen automatically

src/TimeSeries/index.js
mop9047: I agree! these were a copied block from the DIY NN class, I will add a catch-all assuming that regression is the default, although maybe there is a discussion needed to decide otherwise. For the DIY NN class, I see that the default task is set to null. I could also add a condition to check if the task remains null and send an error instead.

src/TimeSeries/timeSeriesData.js
mop9047: Just removed this comment as it was only helpful for a specific error that I already fixed.

src/TimeSeries/timeSeriesData.js
mop9047: done :)

examples/timeSeries-load-model-hand-gestures/model/model.json
mop9047: done :)

src/TimeSeries/index.js
mop9047: done :)

src/TimeSeries/index.js
mop9047: I agree, I'll leave this comment unresolved in this case, so it doesn't get forgotten

src/TimeSeries/timeSeriesData.js
mop9047: this line was in the DIY NN class, I tested without it and there nothing changed or any errors so I will probably remove.

examples/timeSeries-weather-prediction/sketch.js
```diff
@@ -0,0 +1,183 @@
...
+// loadData() cannot yet be used as it is formatted differently
+function preload() {
+  json_data = loadJSON("weather_data.json");
+
+  // set the options to initialize timeSeries Neural Network
+  let options = {
+    task: "regression",
+    dataMode: "linear",
+    debug: "true",
+    learningRate: 0.01,
```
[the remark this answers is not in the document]
mop9047: done :)

examples/timeSeries-weather-prediction/sketch.js
mop9047: You're right! it was unnecessary. However, for this example, I changed the output to be the same variable targets which acts both as defining which will be included as the outputs and in the sliding window helper function. However in this case, I used all features as targets to create a loop for creating continuous data.

mop9047:
> > Some additional comments along the lines of @gohai's review! Also, I see this PR still includes `ml5.timeSeries()`, my memory is shaky of course, but I recall a discussion around this implementation being a part of `ml5.neuralNetwork()` with a new "time series" task? Is that correct?
> 
> My memory is also shaky, @shiffman - but I like the idea. Since the options are being passed to the factory method, we could _keep_ the inheritance model that Mathew built, and have `ml5.neuralNetwork()` simply instantiate the specialized class whenever encounters a related `task` option. The alternative is to integrate Mathew's work even more tightly into the existing NeuralNetwork classes. Do you have a preference @mop9047?

Yes! I do remember having that discussion, but I don't clearly remember if we had a decision on which method to do. I definitely think its worth trying out and I have created a new branch to test it [here](https://github.com/ml5js/ml5-next-gen/tree/timeseries-as-task). 

However, there are some changes this method necessitates: I have tried implementing DIY time series and DIY neural network classes and the factory functions in the neural network factory function so it is all under one neuralNetwork call. However since the DIY time series class depends on an import of the DIY neural network class, it is not possible to call the DIY time series class inside of the neural network factory function as it needs to import the DIY ts class in the neural network file. This causes a circular dependency as imports of the other files are present. Thus, in the initial commit on the branch I provided, removed the neuralNetwork factory function from the neural network file and moved it all the way to the time series file to prevent this conflict. So far, all of the examples have worked with this method. 

I have yet to try another method where there is a separate file for the neural network factory function, so there is no circular dependencies, however I am not sure where in the file system to put it in and a good naming convention for it. Let me know if you need any more clarification for this!

examples/timeSeries-hand-gestures/sketch.js
mop9047: I have removed the console logs! and for this one I simplified the example even further, in the latest commit I refactored the example code to simply use "gesture #1" and "gesture #2" instead. I also made the controls similar to the Mouse Gesture example to keep things consistent and more simple. Let me know if you still want to keep the user inputted ones.

examples/timeSeries-load-model-hand-gestures/model/model_meta.json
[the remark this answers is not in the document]
mop9047: Done :)

src/NeuralNetwork/Sequential/index.js
```diff
@@ -0,0 +1,210 @@
...
+      this.neuralNetworkData.meta.seriesShape,
+      this.options.hiddenUnits,
+      this.numberOfClasses // For output units if needed
+    );
+
+    const task = this.options.task;
+    let taskConditions = task;
+    switch (taskConditions) {
+      case "sequenceRegression":
+        layers = tsLayers.regression;
```
gohai, line 100: Best to keep the name in tsLayers exactly the same as the task, so that searching for e.g. "sequenceRegression" will quickly bring up the layer definition

examples/neuralNetwork-sequence-hand-gestures/sketch.js
```diff
@@ -0,0 +1,180 @@
...
+let hands = [];
+
+let sequence = [];
+let targetLength = 30;
+
+let gestures = ["Gesture #1", "Gesture #2"];
+let counts = { "Gesture #1": 0, "Gesture #2": 0 };
+
+let state = "collection";
+let currGesture = gestures[0]; //set currGesture to gesture 1 by default
```
gohai, line 20: (OCD-fueled comment:) The comments in your code seem to oscillate between capitalizing the first word, and not doing it, without much apparent reason. Best to figure out _a_ rule, and stick to it ;)
(And always a space after `//`, please)

PimTournaye:
Hi all, I was talking with @gohai about ml5 yesterday, and the subject of audio came up. It was mentioned that this PR was being discussed. There are other libraries out there that can help with audio and machine learning, but it would be nice to also have this capability in ml5, which currently supports pre-trained audio models but doesn't allow users to train custom audio classifiers.

Should audio be considered within this time-series implementation, or would it warrant a separate audio-specific module, or even an adapter to something like [essentia.js](https://github.com/MTG/essentia.js)?

examples/neuralNetwork-sequence-weather-prediction/index.html
```diff
@@ -0,0 +1,33 @@
+<!--
+  👋 Hello! This is an ml5.js example made and shared with ❤️.
+  Learn more about the ml5.js project: https://ml5js.org/
+  ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
+ 
+  This example demonstrates Training a Weather Predictor through ml5.neuralNetwork with sequeceRegression Task.
```
gohai, line 6: @mop9047 I enjoyed the simplicity of the example. For making the visual output more interesting: how about a bar in a bar-graph gets drawn per hour predicted? (Since you know the training data, you wouldn't need to deal with automatically scaling the y-axis, and instead used some fixed min/max.)
You could divide the width of the canvas into 24 equally-spaced columns, and draw one by one, as the user predicts one hour after the next.

examples/neuralNetwork-sequence-train-mouse-gesture RDP/sketch.js
```diff
@@ -0,0 +1,132 @@
+/*
+ * 👋 Hello! This is an ml5.js example made and shared with ❤️.
+ * Learn more about the ml5.js project: https://ml5js.org/
+ * ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
+ *
+ * This example demonstrates How to train your own mouse gesture classifier through ml5.neuralNetwork with sequeceClassification Task.
```
gohai, line 6: @mop9047 The `train` in the directory name seems superfluous (it's quite long already), and I'd use a dash before `RDP` (since spaces in filenames are at times an issue). Perhaps `neuralNetwork-sequence-mouse-gesture-rdp`?

examples/neuralNetwork-sequence-weather-prediction/index.html
```diff
@@ -0,0 +1,33 @@
...
+  <body>
+    <script src="sketch.js"></script>
+
+    <div style="position: absolute; margin: 350px 0px 0px 100px">
+      <button id="pred_but">Predict Next Hour</button>
+    </div>
+  </body>
+
+  <style>
+    button {
```
gohai, line 28: I'd keep whatever the default styling for buttons is

examples/neuralNetwork-sequence-train-mouse-gesture RDP/index.html
```diff
@@ -0,0 +1,37 @@
...
+
+  <body>
+    <div id="canvasDiv">
+      <script src="sketch.js"></script>
+    </div>
+
+    <div>
+      <button id="recCircle">Record Circle</button>
+      <button id="recSquare">Record Square</button>
+      <button id="trainBut">Train & Predict</button>
```
gohai, line 27: "Finished"?

examples/neuralNetwork-sequence-train-mouse-gesture RDP/sketch.js
```diff
@@ -0,0 +1,132 @@
...
+  recCircle.mouseClicked(recordCircle);
+  recSquare.mouseClicked(recordSquare);
+  trainBut.mouseClicked(trainModel);
+
+  function recordCircle() {
+    state = "collection";
+    currShape = "circle";
+
+    background(220);
+    text(state + " : " + currShape, 50, 50);
```
gohai, line 114: It'd be nice if this text already showed up when the sketch launched. "Currently recording: "?

examples/neuralNetwork-sequence-train-mouse-gesture RDP/sketch.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+}
+
+// code to signify drawing can be done again
+function mouseReleased() {
+  if (mouseY < height && mouseX < width) {
+    // if state is collection, add whole sequence as X, and shape as Y
+    if (state == "collection") {
+      let target = { label: currShape };
+      let paddedCoordinates = model.padCoordinates(sequence, targetLength);
```
gohai, line 52: `padCoordinates` runs RDP internally, right? Perhaps you could think of alternative names for this method as options... ("padding" sounds a bit like there's only things added, but if I understand RDP correctly, this will also simplify/decimate if needed?)

Since the algorithm is quite interesting in itself, you might also consider running this continuously, e.g. inside `draw()` and visualize the resulting coordinates it comes up with (e.g. by drawing small squares around the points). That might look quite interesting, and tell our audience a bit more about what's going on.

And lastly: does `padCoordinates()` have some expectation about the input data (e.g. does it assume an `x` and `y`?).

examples/neuralNetwork-sequence-load-model-hand-gestures/index.html
```diff
@@ -0,0 +1,45 @@
...
+ 
+  This example demonstrates loading a Sign Language classifier through ml5.neuralNetwork with sequeceClassification Task.
+-->
+
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>ml5.js Time Series Hand Gesture load model</title>
```
gohai, line 14: I would flip the filename, so that it's clear that this is a further exploration of `hand-gestures`. So: `neuralNetwork-sequence-hand-gestures-load-model`

examples/neuralNetwork-sequence-hand-gestures/index.html
```diff
@@ -0,0 +1,42 @@
+<!--
```
gohai, line 1: Nitpick: `hand-gestures` is plural while `mouse-gesture` is singular. (Maybe singular throughout?)

examples/neuralNetwork-sequence-load-model-hand-gestures/index.html
```diff
@@ -0,0 +1,45 @@
...
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.9.4/p5.min.js"></script>
+    <script src="../../dist/ml5.js"></script>
+  </head>
+
+  <body>
+    <script src="sketch.js"></script>
+    <div id="canvasDiv"></div>
+    <p>
+      This example loads a model that is trained with ASL hand gestures for
```
gohai, line 23: The link to the instructions how to do Hello and Goodbye in ASL (below) is a fantastic touch, but maybe we wouldn't need the text above (including the introduction) ideally?

examples/neuralNetwork-sequence-train-mouse-gesture RDP/index.html
```diff
@@ -0,0 +1,37 @@
...
+    <script src="../../dist/ml5.js"></script>
+  </head>
+
+  <body>
+    <div id="canvasDiv">
+      <script src="sketch.js"></script>
+    </div>
+
+    <div>
+      <button id="recCircle">Record Circle</button>
```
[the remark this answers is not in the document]
gohai: Or "Start collecting circles"

src/NeuralNetwork/Sequential/timeSeriesData.js
```diff
@@ -0,0 +1,362 @@
...
+import axios from "axios";
+import { saveBlob } from "../../utils/io";
+import modelLoader from "../../utils/modelLoader";
+import nnUtils from "../NeuralNetworkUtils";
+
+import tsUtils from "./timeSeriesUtils";
+
+import NeuralNetworkData from "../NeuralNetworkData";
+
+class TimeSeriesData extends NeuralNetworkData {
```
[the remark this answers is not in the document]
mop9047: Done :)

examples/neuralNetwork-sequence-train-mouse-gesture RDP/sketch.js
mop9047: Yes I agree, done :)

examples/neuralNetwork-sequence-weather-prediction/index.html
mop9047: Done :)

examples/neuralNetwork-sequence-hand-gestures/sketch.js
```diff
@@ -0,0 +1,180 @@
...
+  };
+  model = ml5.neuralNetwork(options);
+}
+
+function setup() {
+  let canvas = createCanvas(640, 480);
+  canvas.parent("canvasDiv");
+
+  // setup video capture
+  video = createCapture(VIDEO);
```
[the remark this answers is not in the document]
mop9047: yes I agree, done. Also put a comment for the flipped argument in hand pose.

examples/neuralNetwork-sequence-hand-gestures/index.html
```diff
@@ -0,0 +1,42 @@
...
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.9.4/p5.min.js"></script>
+    <script src="../../dist/ml5.js"></script>
+  </head>
+
+  <body>
+    <script src="sketch.js"></script>
+    <div id="canvasDiv"></div>
+
+    <p>
+      Instructions: <br />
```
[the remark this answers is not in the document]
mop9047: For simplicity sake, I've used the rdp algorithm so it can be any length and no need for the length settings.

I also tried to implement your suggestion but with helpful text. It looks like:

- if no hands in the frame, then the text says put up hands in frame
- if hands are in frame, then it says for which gesture it is recording for
- one training is done, it says put hands in frame again
- if hands are in frame, is says predicting...
- shows prediction once done with

examples/neuralNetwork-sequence-load-model-hand-gestures/index.html
mop9047: Done :)

examples/neuralNetwork-sequence-load-model-hand-gestures/index.html
mop9047: I did the same for the train model from earlier, hopefully it would be intuitive enough to understand.

examples/neuralNetwork-sequence-weather-prediction/sketch.js
```diff
@@ -0,0 +1,155 @@
+/*
+ * 👋 Hello! This is an ml5.js example made and shared with ❤️.
+ * Learn more about the ml5.js project: https://ml5js.org/
+ * ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
+ *
+ * This example demonstrates loading JSON data and training a Weather Predictor
+ * through ml5.neuralNetwork with the sequenceRegression task.
+ * XXX: let us know what this dataset is from/about
```
gohai, line 8: Hi @mop9047. I started cleaning up this example (please make sure to pull the branch before you do future work). Highlighting the parts that need some work here. I'll look at other examples once we're done with this one. 🙌

examples/neuralNetwork-sequence-weather-prediction/sketch.js
```diff
@@ -0,0 +1,155 @@
...
+  // the JSON file has the actual data in a "data" property
+  data = data.data;
+
+  // XXX: the following is quite unclear
+  // XXX: - try naming the function with an action word (verb)
+  // XXX: - explain why we do this
+  // XXX: - the second and the third parameter need to be the same - just pass one then?
+  // XXX: - the last parameter is 10, but the result has 14 entries 🤷‍♂️ (data has 24)
+  // XXX: - we use the result for inputs and outputs, how about naming those
+  // XXX:   properties with those words directly (rather than "sequences", "targets")
```
gohai, line 52: @mop9047

examples/neuralNetwork-sequence-weather-prediction/sketch.js
gohai: Wouldn't it be nicer to look at data of an _actual_ place, one that means something to you? Why not share with us the weather of your home town, or from Brooklyn, or ...? (Should be trivial, but I'll send you a script for US weather stations via email.)

gohai: https://gist.github.com/gohai/dd7d03fb79160817ade763aab33c316c

gohai:
Hi @mop9047!

I tried open-coding the sampling mechanism in the example, see 790a1627a8b12160c75a3098b6d05712e29cb431. Please have a look and tell me if you think this approach is feasible. It also doesn't work anymore (something related to coordinates in `sequentialData.js:188`) - hope you can help me fix it.

* It'd be slightly more ergonomic if your class could simply ignore properties that aren't numbers - then `date` wouldn't need to be unset? 🤔
* Also: aren't we providing the property names for `inputs` and `outputs` in the constructor already? Couldn't it just pick the ones specified? (in which case there'd be even less need to do deep copies of objects to mutate their properties between input and output)

examples/neuralNetwork-sequence-weather-prediction/sketch.js
mop9047: Hi @gohai I agree that would definitely be nicer. I am trying the data from the script you provided but the precipitation is the whole period so the model auto deletes it (I had no idea it had that feature). Can you help me find the list of the weather station id's? I'll try to find a place with a moving precipitation value

examples/neuralNetwork-sequence-weather-prediction/sketch.js
```diff
@@ -0,0 +1,155 @@
...
+  let maxDataValue = 35;
+
+  for (let i = 0; i < graphValues.length && i < maxBars; i++) {
+    let barHeight = map(graphValues[i], 0, maxDataValue, 0, height - 180);
+    let x = i * barWidth;
+    let y = height - barHeight - 20;
+
+    // XXX: the colors are a bit unintuitive (blue is normally associated
+    // with more rain rather than less) - maybe you could do some research
+    // and use lerpColor() here?
```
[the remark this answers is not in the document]
mop9047: Done :)

examples/neuralNetwork-sequence-train-mouse-gesture RDP/sketch.js
mop9047: Hi @gohai, I changed the name to `setFixedLength` let me know if the naming is ok or if we should think this through more

src/NeuralNetwork/Sequential/index.js
mop9047: Done :)

examples/neuralNetwork-sequence-weather-prediction/sketch.js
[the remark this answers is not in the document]
gohai: Looks great now!

gohai:
> Hi @gohai ! I fixed the error, it was because the output also had a length of 10, which made it seq2seq, when the sequenceNN only supports seq2one for now. I changed it so that it takes the multiple samples for the input then take one sample after for the output

Thanks for fixing my problem, @mop9047 - it works well now!

> > * It'd be slightly more ergonomic if your class could simply ignore properties that aren't numbers - then `date` wouldn't need to be unset? 🤔
> Can you clarify what you mean further? when you mean the class do you mean the sequenceNN in general? For now, I believe strings are accepted as inputs since they are also one-hot encoded. It could be useful to have categorical strings in some training use cases.

I see. In this case it's fair to leave them then.

> > * Also: aren't we providing the property names for `inputs` and `outputs` in the constructor already? Couldn't it just pick the ones specified? (in which case there'd be even less need to do deep copies of objects to mutate their properties between input and output)
> For now I think it does throw an error claiming that what is put into the model does not follow the input/output structure defined by the user.

I see some practical upsides in changing it to be "friendly" rather than strict - let me explain:

The user is already defining that his/her inputs shall be `["foo", "bar"]` and the outputs `["foo", "bar"]`. Now if they are passing an _object_ to `nn.addData()` that has some superfluous property `baz`, we could just only pull out values for `foo` and `bar` and leave `baz` alone (without raising an error).

This makes it so that the user wouldn't often need to _clone_ the object in order to achieve what they would want. (The semantics around reference types and shallow vs deep cloning are not always super intuitive for everyone. We'd sidestep this topic by putting more smarts into `addData`.)

examples/neuralNetwork-sequence-weather-prediction/sketch.js
gohai: @mop9047 You could try using the OpenWeatherMap API, which returns you historical data for any `lat`/`long` pair (on the globe). It needs an API key, but supposedly there are a number of free requests per day if you sign up.

```
https://api.openweathermap.org/data/2.5/onecall/timemachine?lat={lat}&lon={lon}&dt={time}&appid={API key}&units=metric
```

Maybe you find some place that has daily rain. Equatorial Rainforests perhaps? (Doesn't this also happen in Indonesia?)

src/NeuralNetwork/Sequential/sequentialUtils.js
```diff
@@ -0,0 +1,764 @@
...
+        obj[label[ind]] = value;
+      });
+      return obj;
+    });
+  }
+
+  convertValueSequence(xInputs, options = null) {
+    const { inputLabels } = options;
+    if (xInputs.length % inputLabels.length !== 0) {
+      throw new error(
```
gohai, line 130: @mop9047 This wants to be upper-case. (Got "error is not defined" when I accidentally passed in an empty sequence.)

examples/neuralNetwork-sequence-mouse-gesture-rdp/sketch.js
```diff
@@ -0,0 +1,126 @@
...
+    line(sequence[i].x, sequence[i].y, sequence[i + 1].x, sequence[i + 1].y);
+  }
+
+  // This sketch uses the RDP line simplification algorithm
+  // to make each inputs to the neural network have the same
+  // number of points.
+  // For more information about RDP, see:
+  // https://www.youtube.com/watch?v=ZCXkvwLxBrA
+
+  let rdp = model.setFixedLength(sequence, targetLength);
```
gohai, line 48: @shiffman @MOQN Wondering if you have some (additional) ideas for how to name this (helper) method, which takes a sequence of indeterminate length and returns a sequence of fixed length - using the RDP algorithm?

gohai: `simplify`? `rdp`? `resample`?

src/NeuralNetwork/Sequential/sequentialUtils.js
mop9047: Thanks for the catch! Have fixed it now, all the other errors in this file are also not capitalized.

examples/neuralNetwork-sequence-hand-gesture/sketch.js
```diff
@@ -0,0 +1,198 @@
+/*
```
gohai, line 1: @mop9047 I reworked this example a little bit, please have a look and let me know if you'd rather do things differently!

examples/neuralNetwork-sequence-hand-gesture/sketch.js
```diff
@@ -0,0 +1,198 @@
...
+    // hands in frame, add to sequence
+    let handpoints = getKeypoints(["Left", "Right"]);
+    sequence.push(handpoints);
+
+    // This uses the RDP line simplification algorithm to make sure each
+    // input to the neural network has the same number of points.
+    // For more information about RDP, see:
+    // https://www.youtube.com/watch?v=ZCXkvwLxBrA
+
+    let rdp = model.setFixedLength(sequence, targetLength);
```
gohai, line 72: @mop9047 How is `setFixedLength` dealing with the inputs here? (For an array of numbers, are they implicitly assumed to be x1, y1, x2, y2, etc.?)

examples/neuralNetwork-sequence-hand-gesture/sketch.js
```diff
@@ -0,0 +1,198 @@
...
+      noStroke();
+      circle(keypoint.x, keypoint.y, 5);
+    }
+  }
+}
+
+// Return the tracked hand points as flattened array of 84 numbers
+// for use as input to the neural network
+
+function getKeypoints(whichHands = ["Left", "Right"]) {
```
gohai, line 163: @mop9047 This was a bit tricky for one reason: handPose can return one or two hands, which means this function would return _either_ 42 or 84 numbers. (I am not sure if the order the hands appear in the array is guaranteed either.)

Currently this now returns always 84 numbers (with zeros for hands that aren't showing), but I am not sure if this is the best way... do you think it makes sense for this to be able to record gestures that make use of both hands? Or, other direction: would it be bad to specify that a specific hand need to be used to make a gesture (and the sketch only looks for that one)?

examples/neuralNetwork-sequence-hand-gesture/sketch.js
mop9047: @gohai yes this is an assumption that I made based on other sample code using the hand pose model. Would there be any case where this would cause any issues?

examples/neuralNetwork-sequence-hand-gesture/sketch.js
mop9047: @gohai Thanks! I think it looks great. The way I designed these sketches was mainly based off of the rock paper scissors example from the NeuralNetwork examples. I personally don't mind straying away from that and I think its also easier without the extra html things in the sketch!

examples/neuralNetwork-sequence-weather-prediction/sketch.js
mop9047: @gohai Not sure if this happens in indonesia, but I did try the monsoon season from where I live in the Philippines and the data just ends up stabilizing at one value. I'm trying to find a place or time that has like a cyclical rain pattern so the predictions look dynamic.

Also in case that you'd want to use it too, I was using a different api in this[ link](https://www.visualcrossing.com/weather-query-builder/#) open weather map is also good but the free tier only offers one piece of data, the one I linked, you can choose as many datapoints as you want.

examples/neuralNetwork-sequence-hand-gesture-load-model/index.html
```diff
@@ -0,0 +1,28 @@
...
+  ml5.neuralNetwork with the sequenceClassificationWithCNN task.
+  This example has been trained with the ASL gestures for Hello and Goodbye.
+-->
+
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>ml5.js neuralNetwork Hand Gesture Loading Pre-trained Model Example</title>
```
gohai, line 16: @mop9047 I slightly reworked the last example - please have a look if anything is amiss 🥳

examples/neuralNetwork-sequence-hand-gesture-load-model/sketch.js
```diff
@@ -0,0 +1,158 @@
...
+ */
+
+let video;
+let handPose;
+let hands = [];
+let model;
+let isModelLoaded = false;
+
+let sequence = [];
+let sequenceLength = 50;
```
gohai, line 22: @mop9047 Change this to 30? If you can, please also upload re-trained model files (with 84 dimensions)

examples/neuralNetwork-sequence-hand-gesture/sketch.js
gohai: No this makes sense. Will just need good documentation ultimately!

src/NeuralNetwork/Sequential/index.js
```diff
@@ -0,0 +1,215 @@
...
+import callCallback from "../../utils/callcallback";
+import setBackend from "../../utils/setBackend";
+
+import seqUtils from "./sequentialUtils";
+import SequentialData from "./sequentialData";
+import { createSeqLayers } from "./seqLayers";
+
+// call an extension of DIY Neural Network as a new class, override select methods
+// which are seen below:
+class DIYSequential extends DiyNeuralNetwork {
```
gohai, line 13: @mop9047 Perhaps as the last task for this period: please try to add some friendly error messages where possible 🙌 The few instances I have encountered:

* training with only one (output) class: `Error in oneHot: depth must be >=2, but it is 1
    at oneHot_` (if this is worth fixing in the regular NN, please open a separate PR!)
* forgetting to call `normalizeData`: `nput 0 is incompatible with layer conv1d_Conv1D1: expected ndim=3, found ndim=2
    at Conv1D.assertInputCompatibility`
* varying the length of the sequence during training: `Error when checking input: expected conv1d_Conv1D1_input to have 3 dimension(s). but got array with shape 0,1
    at standardizeInputData`
* during inference, submitting sequences with length 30 against a model that was trained with 50: `Error when checking : expected conv1d_Conv1D1_input to have shape [null,50,40] but got array with shape [1,30,42].`

examples/neuralNetwork-sequence-hand-gesture-load-model/index.html
```diff
@@ -0,0 +1,28 @@
+<!--
```
gohai, line 1: @shiffman @MOQN *Very unimportant*, but I thought I raise this here before I forget: was wondering if we prefer `neuralNetwork-sequence-...` or a (shorter) `neuralSequence-...`? (Feeling the former is truer to the hierarchy - but maybe too long?)

examples/neuralNetwork-sequence-hand-gesture-load-model/sketch.js
gohai: That's great - thank you!

I think both (showing or not showing the path in this example) could work. Perhaps leaning a bit towards the latter, so that there is more differentiation to what your other hand-gesture example offers, and we keep this a way of demonstrating how a clean, "user-facing" application could look like?

examples/neuralNetwork-sequence-hand-gesture/sketch.js
mop9047: From my own experience, it depends on how complex and how many categories are, in our case of two categories and only some simple gestures we definitely can do lesser points but often in sign language as an example there are similar gestures that from the perspective of the model will look the same.

src/NeuralNetwork/Sequential/index.js
mop9047: @gohai I have created the some friendlier errors for these cases, let me know if the wording of the error is a bit confusing

mop9047: For the first error, it also happens in the base neuralNetwork class so I made PR for the first error here: https://github.com/ml5js/ml5-next-gen/pull/268

gohai: Great - thanks so much @mop9047!

examples/neuralNetwork-sequence-weather-prediction/sketch.js
gohai: Thank you for exploring this!

It's an interesting question: what to do with an example, if it only works with synthetic (toy) data... on one hand it will convey the promise and application of the technique - on the other hand, users will probably try to use it for their own applications/data and then have disappointing results... maybe @shiffman has some thought on this (whenever he has an opportunity to look into this).

I'd love to keep this (or a similar) example around, since it is the only one that actually uses LSTM (rather than CNN). Could you perhaps share the real-life training data you tried with somewhere (gist?), so that we could also play with it?

shiffman: Yes, I agree that we would ideally use a real data set, if it's synthetic I might suggest actually using a math formula and applying randomness to it and then having clear comments and info about how the data was generated? I'm happy to help with gathering data from an API, I'm doing APIs in my class this fall! I love the idea of the data being tied to where you are from @mop9047!

PR #254, opened 2025-06-07
Change to name exports

pearmini:


shiffman:
Hi @pearmini, thank you for working on this, super exciting! Is the main reason for the large size the "bundling" of tensorflow.js within ml5.js? Is that something we should reconsider or discuss or does ml5.js perhaps not require a full tensorflow.js bundle? I ask in anticipation of the possibility of including transformers.js models as well in the future?

pearmini:
Hi @shiffman,

I believe the large bundle size mainly comes from both tensorflow.js and the individual models. For example, the total size of ml5-bodyPose.min.js is around 2.3MB — with tensorflow.js accounting for approximately 1.2MB and the body pose model itself about 1.1MB.

That said, after tree-shaking support is introduced in this PR, bundling in a Node environment has become much more efficient. If users use modern bundlers like Rollup, Webpack, or esbuild, only the models they actually use will be included in the final output. So in those cases, including `transformers.js` models shouldn’t be a major concern for bundle size.

However, for the browser environment, things are more challenging. The current `ml5.min.js` includes full `tensorflow.js`, which is required and limits our ability to reduce size significantly. One feasible solution might be to modularize the library (like we do in this PR) — allowing users to load models via separate bundles and treating `tensorflow.js` and `transformers.js` as external dependencies.

```html
<script src="tensorflow.min.js"></script>
<script src="transformers.min.js"></script>
<script src="ml5-body-pose.min.js"></script>
<script src="ml5-hand-pose.min.js"></script>
<script src="ml5-transformer-model.min.js"></script>
```
This would keep the core bundles lean and give users more flexibility in managing what they load.

I’ll take a closer look at the dependency graph to pinpoint where the size is coming from. Definitely looking forward to discussing this more with you!

gohai:
Posting a thought here (before it escapes me): in previous conversations, @shiffman had imagined a scenario where we might switch between tensorflow.js and transformers.js-backed pretrained models by passing different values to the constructor... e.g. we have already `BodyPose("MoveNet")` and `BodyPose("BlazePose")`, and there might be more (possibly built with transformers). Just wondering if/how this could mesh with your proposal here @pearmini? e.g. is there a conceivable way how this wouldn't pull in _all_ the dependencies regardless? (I don't think we're very set on this functionality either, but it was something that we once imagined - perhaps worth exploring as you weigh the cost/benefits of various directions...)

pearmini:
@limzykenneth Thanks so much for the valuable suggestions!

ml5 is facing a similar issue as p5.js—bundling all the modules takes nearly a minute! Switching to Rollup or Rolldown could be a great improvement, as both significantly reduce bundle time and maybe a little bit bundle size. In that case, migrating to Vite for development might also make sense.

However, for now, Rspack seems like an easier solution since it's more stable and compatible with the current configuration.

I'll experiment with these options if we need them and dive deeper into the dependencies to explore further optimizations!

pearmini:
>  is there a conceivable way how this wouldn't pull in all the dependencies regardless

@gohai This is definitely possible! We could introduce an explicit API pattern like the following:

```js
import {faceMesh, tensorFlow} from "ml5";

const faceMesh = await faceMesh({
  runtime: tensorFlow()
});
```
```js
import {faceMesh, transformer} from "ml5";

const faceMesh = await faceMesh({
  runtime: transformer()
});
```

The key idea is that we pass the runtime engine explicitly to the model constructor, which allows bundlers to include only the engine actually used. However, this also means the runtime is no longer optional—otherwise, the default would be bundled regardless:

```js
import {faceMesh, tensorFlow} from "ml5";

const faceMesh = await faceMesh(); // This will not work as expected.
```

shiffman:
Thanks all for the wonderful thoughts on this thread! I think there is one missing piece to this discussion. There is both the concept of a "task" in ml5.js as well as a "model" selection for that task. 

I believe it makes the most sense to have smaller ml5.js bundles on the "task" level but we could consider the model level. For some tasks there is is only one model , but others might have multiple options.

The selection of tensorflow.js or transformers.js, however, happens at the model level.  So let's take the case of depth estimation from #248, for example, we might have:

```js
// This model is from tensorflow.js
let depthEstimator = await ml5.depthEstimation("ARPortraitDepth");
```

```js
// This model is from transformers.js
let depthEstimator = await ml5.depthEstimation("depth-anything-v2-small");
```

It may simplify things, however, to restrict a task to only one back-end. For depth estimation and object detection, the approach depends on the research of @nasif-co and @yiyujin happening now in parallel!

In the case of, say, bringing in `ml5.speechToText()` we would load the whisper model with transformers.js and there would be no equivalent tensorflow.js option. 

I'd be curious what the bundle size would be including both tensorflow.js and transformers.js?  I worry about requiring the beginner to both add an ml5.js script tag along with the correct tensorflow.js or transformer.js version. That may be confusing and prone to error?

Adding @xenova who may have some thoughts or advice from the transformers.js side!

PR #257, opened 2025-06-26
add legacy object detection cocossd code and example from p5js

yiyujin:
<h3>Bring back object detection in ml5.js</h3>

To start, I added [legacy object detection code](https://github.com/ml5js/ml5-library/tree/main/src/ObjectDetector)) and [p5js example code](https://editor.p5js.org/ml5/sketches/ObjectDetector_COCOSSD_Video). Example code is from [github reference doc](https://github.com/ml5js/ml5-library/blob/main/docs/reference/object-detector.md)

<h3>History</h3>
Make code compatible with ml5 1.0

- load the model in preload() for basic example. (this will change later with p5.js 2.0 integration) -> https://github.com/ml5js/ml5-next-gen/commit/7deb4c87e0cbaa5e6b3fe62580b6a81261aff8cb
- new detectStart() and detectStop() functions (instead of calling detect() recursively over and over. -> https://github.com/ml5js/ml5-next-gen/commit/ce82f51a76278e6f717d199c6aa7fa32c7650c95
- Swap order of error and results argument for results callback. This way the error is optional and handling is not required in basic example. -> https://github.com/ml5js/ml5-next-gen/commit/4db442fab7dae0b36c0d181105bb2fe21734e77d

Make code scalable to support more models
- refactor ObjectDetector Class (ObjectDetector wraps model -> ObjectDetector as model) -> https://github.com/ml5js/ml5-next-gen/commit/9375570ae7c07eb7eadea57fe955c19a58ea2602

<h3>Next Steps</h3>
- p5 2.0 examples : Having this coco-ssd work with p5.js 2.0 is priority over supporting Transformers.js.

| img |
|:--:|
| p5js 2.0 error (although it doesn't halt the detection) |

### DIFF
--- a/examples/objectDetection-webcam/sketch.js
+++ b/examples/objectDetection-webcam/sketch.js
@@ -3,7 +3,7 @@
  * Learn more about the ml5.js project: https://ml5js.org/
  * ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
  *
- * This example demonstrates detecting objects in a live video through ml5.imageClassifier.
+ * This example demonstrates detecting objects in a live video through ml5.objectDetector.
  */
 
 let video;


shiffman:
Yay! Great job @yiyujin! As discussed, if it's helpful here's a quick list of things that we changed with ml5.js 1.0.

* load the model in `preload()` for basic example. (this will change later with p5.js 2.0 integration)
* new `detectStart()` and `detectStop()` functions (instead of calling `detect()` recursively over and over.
* Swap order of `error` and `results` argument for results callback. This way the error is optional and handling is not required in basic example.

yiyujin:
✅ load the model in preload() for basic example. (this will change later with p5.js 2.0 integration)

- [ ] new detectStart() and detectStop() functions (instead of calling detect() recursively over and over.

- [ ] Swap order of error and results argument for results callback. This way the error is optional and handling is not required in basic example.

yiyujin:
Based on today's meetup with @shiffman cc.@nasif-co

After all, final output for the project would be the p5js example code.

Example code (in order of priority) :
- [] object detection on image
- [] object detection on webcam
- [] ( object detection on video - would be interesting for curation purpose. ex. blazePose - was easy to demo with a shot video

I guess next will be the docs, which I can draft out in parallel with the code !

yiyujin:
I made the [3 changes](https://github.com/ml5js/ml5-next-gen/pull/257#issuecomment-3055625569) to code so it works with ml5.js 1.0

💪 This week, I want to digest code better and draft a docs that explains the common pattern (functions and its roles) used across models : start, stop etc. I need it for myself anyways + could be a useful onboarding docs for future dev team.

Some notes on naming conventions : 
Last meeting (w Dan, Nasif) I recall we briefly talked about naming conventions, how tensorflow.js uses a general word "predict" while ml5.js uses more model specific words like classify, detect, etc. I guess it's also different by context : user code vs. source code (such as results vs. detections / classifications..). 
Anyhow, consistency matters so maybe a central docs can include some naming convention info too. Not much opinion on this so far, but just wanted to mention before I forget :-)
 
(speaking of which.. I think my branch should have been object-detector, not object-detection..? 😭)

examples/objectDetection/sketch.js
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2020 ml5
+//
+// This software is released under the MIT License.
+// https://opensource.org/licenses/MIT
+
+/* ===
+ml5 Example
+Object Detection using COCOSSD
+This example uses a callback pattern to create the classifier
+=== */
```
shiffman, line 10: This is a tiny thing, but sometime this year we adopted a "friendlier" pattern for comments at the top of an example sketch. I don't think we need the copyright and technically the license should be the ml5.js one. Here is what it looks like from a hand pose example:

```js
/*
 * 👋 Hello! This is an ml5.js example made and shared with ❤️.
 * Learn more about the ml5.js project: https://ml5js.org/
 * ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
 *
 * This example demonstrates face tracking on live video through ml5.faceMesh.
 */
 ```
 
 (This also reminds me that we discussed moving the Code of Conduct to the website, I forgot where we are with that though maybe @MOQN remembers?)

File: examples/objectDetection/sketch.js [hunk 2]
```diff
@@ -0,0 +1,52 @@
...
+
+function gotDetections(results) {
+  detections = results;
+}
+
+function draw() {
+  image(video, 0, 0);
+
+  for (let i = 0; i < detections.length; i += 1) {
+    const detection = detections[i];
```
shiffman, line 38:
```suggestion
    let detection = detections[i];
```

We're adopting p5.js style of using `let` even where `const` would be more typical JS.

examples/objectDetection/sketch.js
```diff
@@ -0,0 +1,52 @@
...
+  video = createCapture(VIDEO);
+  video.size(width, height);
+  video.hide();
+
+  detector.detectStart(video, gotDetections);
+}
+
+function gotDetections(results) {
+  detections = results;
+}
```
shiffman, line 32: I wouldn't over do it, but a few concise explanatory comments might be good to add.

shiffman (commented):
Hi @yiyujin fantastic work! I left a few small comments in the basic example!

examples/objectDetection/sketch.js
yiyujin: @shiffman thank you for the comments! I will update it for review.

yiyujin:
I updated the three example codes -> bcd1788e1246c6cbb5804342ac8ba2186344ceea

**- webcam** (examples/objectDetection-webcam/sketch.js)
**- single image file** (examples/objectDetection-sigle-image/sketch.js)
**- video file** (examples/objectDetection-video/sketch.js)

I'm esp. not sure if the video file example is at its best, since I'm adding **a bunch of video eventListeners** to check if video is loadedmetadata/play/pause/ended. I don't think p5 has a method on video to do that if I'm correct..

https://github.com/user-attachments/assets/b0a31e21-25db-48dc-a2ce-c28c43fe9f6d

examples/objectDetection-video/sketch.js
```diff
@@ -0,0 +1,89 @@
...
+  video.elt.addEventListener('loadedmetadata', () => {
+    console.log('Video metadata loaded');
+
+    // Resize canvas to match video size
+    resizeCanvas(video.elt.videoWidth, video.elt.videoHeight);
+  });
+  
+  video.elt.addEventListener('play', () => {
+    console.log('Video started playing');
+
```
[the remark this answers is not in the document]
yiyujin: oh okay! I will forget about showControls and try video.hide + video.loop(). Thank you 🙏

examples/objectDetection-video/sketch.js
```diff
@@ -0,0 +1,61 @@
...
+  detections = results;  
+}
+
+function setup() {
+  // Create canvas with initial size - will be resized later
+  createCanvas(640, 480);
+  
+  // Load and loop the video for object detection
+  video = createVideo('test.mov');
+  video.size(width, height);
```
shiffman, line 29: Just noting our discussion to use a video we have permission for. Also, I would suggest sizing the video to 640x480 so it doesn't appear squashed or skewed. Or just make the canvas size identical to the video. You can add a code comment that explains that if the video does not match the canvas size, you may need to adjust the code to scale.

examples/objectDetection-video/sketch.js
```diff
@@ -0,0 +1,61 @@
...
+  let scaleX = width / video.elt.videoWidth;
+  let scaleY = height / video.elt.videoHeight;
+
+  for (let i = 0; i < detections.length; i++) {
+    let detection = detections[i];
+
+    let x = detection.x * scaleX;
+    let y = detection.y * scaleY;
+    let w = detection.width * scaleX;
+    let h = detection.height * scaleY;
```
shiffman, line 49: Noting that if the video is sized the same as the canvas then this is unnecessary which is preferable for a basic example. However, if we do have to use this we should use the p5.js `video.width` and `video.height` property directly.

I would have thought if you call `video.size(640, 480)` then the x, y, width, height returned would match the new size. Could this be related to the bug @nasif-co is fixing in #264??

examples/objectDetection-webcam/sketch.js
```diff
@@ -0,0 +1,53 @@
...
+function preload(){
+  detector = ml5.objectDetector("cocossd");
+}
+
+function setup() {
+  createCanvas(640, 480);
+
+  // Using webcam feed as video input, hiding html element to avoid duplicate with canvas
+  video = createCapture(VIDEO);
+  video.size(width, height);
```
shiffman, line 22: Ah, I just tested and confirm that the #264 applies here too! If I change the webcam to `320,240` then the bounding boxes are all off. @yiyujin I don't think this has to be fixed in this PR, it can be done separately after the fact or along with #264. @nasif-co maybe you can help track this in the issue thread or a new issue so that we can merge at least the SelfieSegmentation fix?

shiffman (commented):
A couple last small comments on the video example and then some discoveries about video resizing! Great work @yiyujin!

yiyujin:
- [ ] create p5 2.0 examples

PR #258, opened 2025-07-07
Async constructor for p5 2.x

pearmini:


gohai:
Thank you @pearmini - this is looking great. Two observations:

- Dan expressed his preference for having the 2.x-style examples be accessible through the web editor in the next year, and I too think that this would be useful to have. In the light of this: how about keeping the examples in the `examples\` folder, but renaming them to `-p5-2.0`? (This way, `scripts/uploadExamples.js` will be able to deal with it.) If you prefer to have a different folder for them, you could also modify the script and still give them the prefixed directory names.
- Typo in one of the commit messages ("Update proload name") - but maybe we'll squash & merge anyway?

webpack.config.js
```diff
@@ -24,7 +24,6 @@ const developmentConfig = {
     publicPath: "/dist/",
   },
   devServer: {
-    port: 8080,
```
gohai, line 27: A unrelated/spurious change?

shiffman (dismissed):
This looks great to me! Excited to merge! I don't have any comments beyond @gohai's. I'll note, however, that it might be necessary to review and test [`uploadExamples.js`](https://github.com/ml5js/ml5-next-gen/blob/main/scripts/uploadExamples.js) along with this PR. I'm not sure if changes will be needed to accomondate uploading both the p5.js 1 and 2 examples.

gohai (dismissed):

gohai:
Thanks @pearmini!

`scripts/updateP5Version.js` might need some changes to deal with the two-versioned nature of the examples. Actually, the script is probably broken as-is, since it fetches the most recent p5 version number (2.0.3) and uses this for _all_ the examples. Not sure if the p5.js team is planning on releasing further 1.x updates this year (or if we'll do so many releases), but the script could walk `https://api.github.com/repos/processing/p5.js/releases` instead, pull out the latest for both, and use it for both.

gohai:
Since the `-p5-2.0` directories will need some special handling, it might make sense to include it here. (Picking the most recent 1.x version, rather than the most recent overall version could be a separate, precursor patch - but probably not worth the extra mile, unless you want to.)

pearmini:
@gohai I've updated scripts/update-p5-version to support multiple p5 versions. Notably, it now fetches versions directly from npm instead of github realeases, making it easier to get the latest release for each major version.

```js
async function getP5VersionsFromNpm() {
  const res = await fetch("https://registry.npmjs.org/p5");
  const data = await res.json();
  return {
    latest: data["dist-tags"].latest, // Latest 2.x version
    r1: data["dist-tags"].r1, // Latest 1.x version
  };
}
```

I also ran the script and updated the p5 1.x version used in the examples.

pearmini:
@gohai @shiffman @limzykenneth I also opened a PR to fix the issue in p5 2.0 when inspecting the examples

pearmini:
@gohai @shiffman I've already created a new script called `scripts/updateExamplesREADME` to update the examples' README.

PR #268, opened 2025-07-31
Add friendlier message for oneHot error

mop9047:
As suggested by @gohai.

When users give only one class for classification, they receive an error message: 
`Error in oneHot: depth must be >=2, but it is 1`
Changed it to:
`🟪 ml5.js classification error: Classification requires at least 2 different classes, but only found 1 class: (some class name). Please add training data with multiple different output values.`

Tested by changing the outputs to only one value. E.g., in the neuralNetwork-mouse-gesture example, change every output to "red-ish" to see the error.


shiffman:
We're going to merge this in quickly now to sneak it in to the practice release @pearmini and I are working on! @gohai we can adjust this later if you have comments!

PR #272, opened 2025-07-31
Remove documentation.md

gohai:
This appear to have served only as temporary API reference during the Alpha release period of the reconstituted library. Delete it now that the website exist, since we will likely not remember to update this file as well (and we don't want to confuse the LLMs with diverging information :)


shiffman (commented):
Thanks for catching this @gohai!

PR #273, opened 2025-07-31
Promote linking to ml5@1 in README

gohai:
This aligns things with how we currently suggest to do it on the website.


shiffman (approved):
This looks great to me! Just want to double check that these changes don't affect the update README script, @pearmini can you confirm?

pearmini (approved):
@shiffman LGTM!

PR #289, opened 2025-11-13
Add Transformer.js as a backend for image classification

pearmini:
For single image:

```js
function preload() {
  classifier = ml5.imageClassifier("VisionTransformer");
  img = loadImage("images/bird.jpg");
}

function setup() {
  createCanvas(400, 400);
  classifier.classify(img, gotResult);
  image(img, 0, 0, width, height);
}
```

For webcam:

```js
function preload() {
  classifier = ml5.imageClassifier("VisionTransformer");
}

function setup() {
  createCanvas(640, 480);
  video = createCapture(VIDEO);
  video.size(640, 480);
  video.hide();
  classifier.classifyStart(video, gotResult);
}
```

Specify options:

```js
function preload() {
  classifier = ml5.imageClassifier("VisionTransformer", {dtype: "32", device: "wasm", topK: 5});
}
```

## TODO

- [x] Classify single image
- [x] Handle Video Input
- [x] Add more examples
- [x] Support top_k input
- [ ] Docs


examples/imageClassifier-transformer-webcam/index.html
```diff
@@ -0,0 +1,22 @@
+<!--
+  👋 Hello! This is an ml5.js example made and shared with ❤️.
+  Learn more about the ml5.js project: https://ml5js.org/
+  ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
+
+  This example demonstrates detecting objects in a live video through ml5.imageClassifier.
```
gohai: add `using a transformer model` (here and in `<title>`, sketch.js)?

examples/imageClassifier-transformer-single-image/index.html
```diff
@@ -0,0 +1,23 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>ml5.js imageClassifier Transformer Example</title>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.11.10/p5.min.js"></script>
+    <script src="../../dist/ml5.js"></script>
+  </head>
+  <body>
+    <script src="sketch.js"></script>
+  </body>
+</html>
+
```
gohai: nitpick: extra newline

nasif-co:
> In my Chrome, I received a warning message and two error messages in the console while it ran. Ideally, we don't print those, since people using the library might think those are their fault, or something they might need to address..? Is there any way to reduce transformers.js verbosity, or otherwise filtering its output?
> 
> Screenshot 2025-12-24 at 5 33 41 PM

These console warnings seem to be an ongoing problem with the onnxruntime itself, here's a github issue on the matter: https://github.com/huggingface/transformers.js/issues/270

xenova:
Hi all :) If you upgrade to Transformers.js v4 (preview)
```
npm i @huggingface/transformers@next
```

these warnings are no longer shown! 🙏

gohai:
Many thanks to thinking of us with this @xenova 🤗 We'll try v4!

colemanliyah:
Hi all, I'm taking over addressing these comments for the summer! One initial question, @gohai I can update the package.json and yarn.lock file with this `npm i @huggingface/transformers@next` to get rid of the errors and then push those changes to github but I just want to confirm that's the preferred route. I know pushing package.json, yarn.lock, ect files can be a slippery situation

src/ImageClassifier/transformer.js
```diff
@@ -0,0 +1,124 @@
...
+    this.needToStop = false;
+
+    // Call the classification loop
+    const next = () => {
+      if (this.needToStop) return;
+      this.classify(inputNumOrCallback, numOrCallback, cb).then(() => {
+        // WebGPU is very fast, so we can call the next frame immediately
+        if (this.device === "webgpu") next();
+        // Wasm is slower, so we wait for 1 second before calling the next frame
+        else setTimeout(next, 1000);
```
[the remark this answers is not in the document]
colemanliyah: @gohai Thanks for pointing this out! You're right, hardcoding the 1000ms setTimeout artificially caps WASM at 1 FPS regardless of the machine's speed.

I can update this to use requestAnimationFrame(next) after the classify() promise resolves. Since classify() is asynchronous, it will naturally wait for the current frame to complete before scheduling the next one without imposing an arbitrary 1-second delay. Is that preferred?

examples/imageClassifier-transformer-single-image-topk/sketch.js
```diff
@@ -0,0 +1,45 @@
+/*
```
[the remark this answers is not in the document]
shiffman: Yes, agreed I think we can remove this example! Good to make sure it works, and `topk` can just live in the documentation!

examples/imageClassifier-transformer-single-image/sketch.js
```diff
@@ -0,0 +1,45 @@
...
+
+// A variable to hold the image we want to classify
+let img;
+
+// Variables for displaying the results on the canvas
+let label = "";
+let confidence = "";
+
+function preload() {
+  classifier = ml5.imageClassifier("VisionTransformer");
```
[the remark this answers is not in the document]
shiffman: I've been thinking about this, basically, this is what the the model name indicates:

* vit = Vision Transformer
* base = the model size tier
* patch16 = the image is divided into 16×16 pixel patches
* 224 = expected input size 224×224 pixels

We can possibly eliminate "patch16" and "224" I think? Maybe for it to be similar to "MobileNet", we use "ViTBase" as the name? Or "VisionTransformerBase". @colemanliyah and I discussed this morning the possibility of finding one other transformers.js image classification model that we could support. If we had that maybe that would help use decide on the naming? Perhaps other models will be different enough that "VisionTransformer" is sufficient.

shiffman:
> Hi all, I'm taking over addressing these comments for the summer! One initial question, @gohai I can update the package.json and yarn.lock file with this `npm i @huggingface/transformers@next` to get rid of the errors and then push those changes to github but I just want to confirm that's the preferred route. I know pushing package.json, yarn.lock, ect files can be a slippery situation

Both package.json and yarn.lock are already updated here so it's ok to commit!

Also, maybe it doesn't matter but we're using `yarn` here instead of `npm`. And Transformers.js v4 has now been released as a stable version, so no need for the next tag! I believe it should be:

```bash
yarn add @huggingface/transformers@^4.2.0
```

examples/imageClassifier-transformer-single-image/sketch.js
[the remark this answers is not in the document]
shiffman: (The reason why I think including "base" is important is that after doing a little bit of research I see a lot of image classification models that are fine-tuned off of this base. For example, if you look at [FloraSense](https://huggingface.co/onnx-community/FloraSense-ONNX)'s model card you'll see:

> fine-tuned Vision Transformer (ViT) model designed for accurate classification of plant species and flora-related imagery. It builds on top of the powerful google/vit-base-patch16-224 base model and is fine-tuned on the Planter_GARDEN_EDITION dataset curated by [Sisigoks](https://huggingface.co/Sisigoks), which includes over 10,000 diverse plant images.

src/ImageClassifier/transformer.js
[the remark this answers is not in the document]
gohai: Thank you for answering my question @colemanliyah! 🙇‍♂️

If `requestAnimationFrame(next)` allows the rest of the sketch to make enough progress, then I think this option (which mirrors the WebGPU behavior as closely as possible) would be the most natural. @pearmini, did you find that it needed a longer delay?

examples/imageClassifier-transformer-single-image/sketch.js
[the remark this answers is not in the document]
gohai: (Love the idea of fine tunes, especially the `Planter_GARDEN_EDITION`! @shiffman)

"VisionTransformerBase" or "ViTBase" both strike me as good / future-proof names. (If this ends up remaining the sole option, we could always add a more general "VisionTransformer" alias later, and existing sketches would work unmodified - going the other direction has more ambiguity which model "VisionTransformer" would refer to under the hood.)

Thank you for looking into this matter (much more deeply then I did 🪴 )

src/ImageClassifier/transformer.js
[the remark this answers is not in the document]
colemanliyah: Okay, this change works with the examples, so I'll keep it in there!

examples/imageClassifier-transformer-single-image/sketch.js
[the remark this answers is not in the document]
colemanliyah: Hi @gohai and @shiffman,

Adding to the conversation here. I did some research and I would suggesst adding either:

1. [ResNet-18](https://huggingface.co/microsoft/resnet-18): Uses convolutional filters to detect local features like edges and textures. It's fast and lightweight in the browser but captures less global context than transformer-based models.
2. [Swin-Tiny](https://huggingface.co/Xenova/swin-tiny-patch4-window7-224): A hierarchical vision transformer that attends to local windows before combining them into larger representations. It offers better accuracy than ResNet while being more efficient than a standard ViT.

They are all trained using ImageNet. I saw this as a pro for our use case because it means they can immediately identify 1,000 common everyday objects right out of the box for beginners. 

Upon looking further into it, I couldn't find other high quality, transfomer.js Image Classification models we could use that were not trained on ImageNet. But please let me know if you guys know any I may have looked over.

As for naming, I do like the ModelNameSize format suggested (e.g., ResNet18 and SwinTiny). My only reservation is that size naming isn't standardized—18, Tiny, Base, etc.—which might be a little confusing for users. Since these are already different architectures, maybe we could use just MobileNet, ViT, Swin, and ResNet? What do you guys think?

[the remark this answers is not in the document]
shiffman: I think it could be good for users to learn about how models come in different sizes and the "18" doesn't bother me too much. We could also maybe have our cake and eat it too by having `Swin` load a default size (probably the smallest) but allow for `SwinTiny` and `SwinLarge` as well? @nasif-co did this ever come up in your work with the `DepthAnything` models? I know those come in different sizes.

One caveat is that for models like `MoveNet` different versions of the same model are selected through a `modelType` property: `MULTIPOSE_LIGHTNING`, `SINGLEPOSE_LIGHTNING`, `SINGLEPOSE_THUNDER`. So we could consider something like:

```js
let classifier = ml5.imageClassifier("Swin", { size: "tiny" });
```

But this doesn't map onto how transformers.js works so maybe the following is best:

```js
let classifier = ml5.imageClassifier("Swin");      // picks tiny by default
let classifier = ml5.imageClassifier("SwinTiny");  // specifies tiny
let classifier = ml5.imageClassifier("SwinLarge"); // specifies large
```

Regarding which models to offer, there are also transformers.js models that were pretrained on ImageNet and then fine-tuned for more specific classification tasksImageNet like the [FloraSense](https://huggingface.co/onnx-community/FloraSense-ONNX) one I referenced above. Here's also a [food classification](https://huggingface.co/onnx-community/swin-finetuned-food101-ONNX) model fine-tuned on Swin. We probably want to be careful about including too many options but perhaps it would be nice to offer at least one model with a different set of classes from the standard 1,000 ImageNet classes?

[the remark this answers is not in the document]
colemanliyah: Ah, I see what you mean. I do like that function calling you suggested with those naming conventions. And I see what you mean about the dataset... in that case, I’d say going with the food model might be more interesting since it’s fine-tuned from a Swin base model, which would give us a different architecture from our default ViT model.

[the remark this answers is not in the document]
colemanliyah: @shiffman I added the food transformer model as the second model for this pr!

src/ImageClassifier/transformer.js
shiffman: I think we should go with names that are closer to the model names on HuggingFace. We could also try @gohai's suggestion of when the model is loaded, console logging the full HuggingFace name.

So maybe `ViTBase` for VisionTransformer and `SwinFood101` for the food classifier. If these are too cryptic we could go with `VisionTransformer` and `Food101`?

shiffman (commented):
One question I have is whether we need these additional `transformer-single-image` and `transformer-webcam` examples. Are they identical to the other `single-image` and `webcam` examples? If so, rather than duplicate the examples, I would suggest simply adding a comment about trying the alternative model. What do you think @colemanliyah? Since the food model is completely different, it's nice to have a separate example for that.

shiffman:
Adding a comment here with some clarified thinking:

These are the examples I suggest we have:

```
imageClassifier-single-image - with default MobileNet, comment about support for other models
imageClassifier-teachable-machine - as is
imageClassifier-webcam - with default MobileNet, comment about support for other models
imageClassifier-food - example demonstrating custom fine-tuned food model
imageClassifier-doodlenet - example demonstrating custom trained doodle model
```

For the doodle example here's an old one from my IMA class that could be adapted: https://editor.p5js.org/ima_ml/sketches/IbXlN6voN

The changes it needs are:
- update to ml5@1
- update to p5@2
- use await/async
- use classifyStart

Open to other comments and feedback!

colemanliyah:
> Adding a comment here with some clarified thinking:
> 
> These are the examples I suggest we have:
> 
> ```
> imageClassifier-single-image - with default MobileNet, comment about support for other models
> imageClassifier-teachable-machine - as is
> imageClassifier-webcam - with default MobileNet, comment about support for other models
> imageClassifier-food - example demonstrating custom fine-tuned food model
> imageClassifier-doodlenet - example demonstrating custom trained doodle model
> ```
> 
> For the doodle example here's an old one from my IMA class that could be adapted: https://editor.p5js.org/ima_ml/sketches/IbXlN6voN
> 
> The changes it needs are:
> 
> * update to ml5@1
> * update to p5@2
> * use await/async
> * use classifyStart
> 
> Open to other comments and feedback!

Since this PR is centered around adding the two transformer models, do you think it would be good to wait to add the doodlenet model?

shiffman:
> Since this PR is centered around adding the two transformer models, do you think it would be good to wait to add the doodlenet model?

Oh, interesting, I hadn't considered that, it's a great point. Yes, adding a DoodleNet example could absolutely be done in a different branch / separately. Either way is ok by me.

colemanliyah:
> > Since this PR is centered around adding the two transformer models, do you think it would be good to wait to add the doodlenet model?
> 
> Oh, interesting, I hadn't considered that, it's a great point. Yes, adding a DoodleNet example could absolutely be done in a different branch / separately. Either way is ok by me.

Okay! Finished reorganizing the examples, aside from adding DoodleNet!

examples/imageClassifier-food/sketch.js
```diff
@@ -0,0 +1,44 @@
...
+
+// A variable to hold the image we want to classify
+let img;
+
+// Variables for displaying the results on the canvas
+let label = "";
+let confidence = "";
+
+async function setup() {
+  classifier = await ml5.imageClassifier("SwinFood101");
```
shiffman, line 22: Maybe a comment here on where the model comes from and how it was fine-tuned on a custom dataset, etc.

examples/imageClassifier-single-image/sketch.js
```diff
@@ -4,6 +4,7 @@
  * ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
  *
  * This example demonstrates detecting objects in an image through ml5.imageClassifier.
+ * Try this example with one of the transformer models as well! "ViTBase" and "SwinFood101"
```
shiffman: Maybe we shouldn't suggest the food one here since it's a bird image although maybe that's a nice learning opportunity to keep it and see what food it thinks a bird is? Regardless, I think it would be useful to explain in the comments that MobileNet and ViTBase both use the same 1,000 labels from ImageNet.

examples/imageClassifier-single-image/sketch.js
```diff
@@ -21,11 +22,11 @@ async function setup() {
   img = await loadImage("images/bird.jpg");
 
   createCanvas(400, 400);
-  
+
   // Initialize the classifier and load the image asynchronously
   classifier = await ml5.imageClassifier("MobileNet");
```
shiffman, line 28: Not sure if the comment about different models would be better here?

shiffman (commented):
A few last comments, after these are addressed, we could have a review from @pearmini and @gohai and then I think it's ready to merge!

shiffman:
> A few last comments, after these are addressed, we could have a review from @pearmini and @gohai and then I think it's ready to merge!

Alas, b/c @pearmini originated this PR, I have to request his review manually here in a comment :)

So excited to see this so close to the finish line! We'll be able to tackle @nasif-co's work on Depth Estimation next!

examples/imageClassifier-single-image/sketch.js
colemanliyah: I think it is best to take it out!

examples/imageClassifier-single-image/sketch.js
colemanliyah: I put it in both places!

examples/imageClassifier-food/sketch.js
```diff
@@ -0,0 +1,46 @@
...
+// A variable to hold the image we want to classify
+let img;
+
+// Variables for displaying the results on the canvas
+let label = "";
+let confidence = "";
+
+async function setup() {
+  // SwinFood101 is a hugging face, image classification model
+  // Trained on a  custom dataset of 101 classes of food https://huggingface.co/datasets/ethz/food101
```
shiffman:
```suggestion
  // SwinFood101 is an image classification model based on the Swin architecture.
  // It was fine-tuned on a custom dataset to recognize 101 categories of food. https://huggingface.co/datasets/ethz/food101
```

examples/imageClassifier-food/sketch.js
```diff
@@ -0,0 +1,46 @@
...
+
+  createCanvas(400, 400);
+  classifier.classify(img, gotResult);
+  image(img, 0, 0, width, height);
+}
+
+// Callback function for when classification has finished
+function gotResult(results) {
+  // The results are in an array ordered by score/confidence
+  // Hugging Face transformers use 'score' instead of 'confidence'
```
shiffman:
```suggestion
  // The results are in an array ordered by confidence
```

I think it's confusing to mention this since ml5.js uses confidence across the board.

examples/imageClassifier-single-image/sketch.js
```diff
@@ -4,6 +4,7 @@
  * ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
  *
  * This example demonstrates detecting objects in an image through ml5.imageClassifier.
+ * Try this example with the transformer model, "ViTBase", it's trained on the same 1000 images as MobileNet
```
shiffman:
```suggestion
* This example demonstrates image classification with ml5.imageClassifier.
* Try this example with the transformer model "ViTBase", which is trained to recognize the same 1,000 ImageNet labels as MobileNet.
```

examples/imageClassifier-webcam/sketch.js
```diff
@@ -4,6 +4,7 @@
  * ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
  *
  * This example demonstrates detecting objects in a live video through ml5.imageClassifier.
+ * Try this example with the transformer model, "ViTBase", it's trained on the same 1000 images as MobileNet
```
shiffman:
```suggestion
* This example demonstrates image classification on a live video with ml5.imageClassifier.
* Try this example with the transformer model "ViTBase", which is trained to recognize the same 1,000 ImageNet labels as MobileNet.
```

shiffman (commented):
Going over this one more time, I made some picky changes to the comments in the examples to be more clear. I'll commit these to save some time!

examples/imageClassifier-single-image/sketch.js
```diff
@@ -21,11 +22,12 @@ async function setup() {
   img = await loadImage("images/bird.jpg");
 
   createCanvas(400, 400);
-  
+
   // Initialize the classifier and load the image asynchronously
+  // Try with a transformer model Replace "MobileNet" -> "ViTBase"
   classifier = await ml5.imageClassifier("MobileNet");
```
gohai, line 28: I am a bit confused why the example loads the `imageClassifier`, and image, twice - first on line 21 and then here, and then again here on line 26 🤔

examples/imageClassifier-food/sketch.js
```diff
@@ -0,0 +1,45 @@
...
+
+// Callback function for when classification has finished
+function gotResult(results) {
+  // The results are in an array ordered by confidence
+
+  console.log("The default results length is 3");
+  console.log(results);
+
+  // Display the results on the canvas
+  fill(255);
```
gohai, line 38: In most other examples, we merely update a global variable in the callback function itself and do the drawing operations inside draw.

I don't know enough about p5.js internals to confidently say whether there are any circumstances where the callback could be called at a time when it _wasn't_ appropriate to draw. When working with students, I probably encourage them more to organize their drawing around draw().

examples/imageClassifier-single-image/sketch.js
```diff
@@ -21,11 +22,12 @@ async function setup() {
   img = await loadImage("images/bird.jpg");
 
   createCanvas(400, 400);
-  
+
   // Initialize the classifier and load the image asynchronously
+  // Try with a transformer model Replace "MobileNet" -> "ViTBase"
```
gohai, line 27: Perhaps: `"Try with a transformer model: replace "MobileNet" -> "ViTBase"`

gohai (commented):
Thank you for continuing this project, @colemanliyah! I only tested the examples, but they worked well for me (Chrome on macOS).

I noticed that for all transformers.js models, the browser sends a request for `tokenizer_config.json`, which fails with a 404. Curious if there’s an easy way to turn this off?

colemanliyah:
> Thank you for continuing this project, @colemanliyah! I only tested the examples, but they worked well for me (Chrome on macOS).
> 
> I noticed that for all transformers.js models, the browser sends a request for `tokenizer_config.json`, which fails with a 404. Curious if there’s an easy way to turn this off?

Hi @gohai Thank you for your comments! As far as this 404 goes, are you able to reproduce it for me? Looks like on my end I don't see this request made. But I can look into it more if needed. Most likely, we would need to have this file in order for it not to error out. My assumption is that Transformers.js just tries to look for this file automatically. I'm not sure if there would be a manual way to turn it off.

Screenshot 2026-09-08 114655

gohai:
@colemanliyah I still see the console output on the current `image-classifier-transformer-js` branch and e.g. the `imageClassifier-webcam` example. (I did run `yarn` before `yarn start`)

Screenshot 2026-09-09 at 5 25 46 PM

When I disable "Network messages" the 404 line disappears (not sure if this enabled by default, but it might be).

Hope this helps! Please let me know if there is anything else I could do or test.

pearmini:
@shiffman @gohai @colemanliyah I just reviewed this PR! I have to say, it’s so good to see how this feature has taken shape! I don't have too much to say about the implementation and the API design. I just have one thought: would it be necessary to include more examples for this new feature? Right now I just saw the example classifying one image. I think it might be helpful if can have more examples to increase the "test coverage", such as handle webcam, a different model, etc,.

shiffman:
Thanks for the feedback @pearmini. @colemanliyah and I discussed this and came to the conclusion that since using `ViTBase` is just a matter of changing one line of code (and its labels are the same 1,000 ImageNet ones) it makes morse sense to add a comment in the existing examples rather than duplicate all of the code.

```js
  // Try with a transformer model. Replace "MobileNet" -> "ViTBase"
  classifier = await ml5.imageClassifier("MobileNet");
```

The "food" model is an example of a model that is fine-tuned and has different labels so we created a separate example for that one. I think we could incorporate more examples in the future, and would be glad to brainstorm ideas for things we would want to demonstrate that are different from these base examples. I think that could happen separate from the PR.

pearmini:
@shiffman Sounds good to me! I think a future PR would be a great idea. This PR is ready to go!

xenova:
Hi! :) Just an FYI before you merge, last week we released Transformers.js v4.3.0, which should fix the webpack issues you identified. We also upgraded ONNX Runtime, so you should see performance improvements there too.
Let me know if you run into any issues!

shiffman:
I'm about to merge this, but I decided to first look into the 404 tokenizer questions that @gohai raised.

It looks like `pipeline()` assumes that the `image-classification` task is `'multimodal'`, so it looks for a tokenizer by default. The models we are using here (`onnx-community/swin-finetuned-food101-ONNX` `Xenova/vit-base-patch16-224`) are vision only and don't have a tokenizer, so Hugging Face returns a 404.

One fix would be to build the pipeline ourselves instead of calling `pipeline()`. 

```js
import {
  AutoModelForImageClassification,
  AutoProcessor,
  ImageClassificationPipeline,
} from "@huggingface/transformers";

const [model, processor] = await Promise.all([
  AutoModelForImageClassification.from_pretrained(hfModelId, { device }),
  AutoProcessor.from_pretrained(hfModelId, { device }),
]);
const classifier = new ImageClassificationPipeline({
  task: "image-classification",
  model,
  processor,
});
```

But maybe this would introduce other issues or make the code less maintainable in the future?

@xenova is my assessment correct? Would it make sense for transformers.js to skip the tokenizer check for vision-only models?

PR #301, opened 2026-03-03
[WIP] Add MobileNet Feature Extractor

JunhaoZhu0220:
This PR implements the `FeatureExtractor` module, adapted from the [ml5.js library (v0.12.2)](https://github.com/ml5js/ml5-library/tree/main/src/FeatureExtractor). It follows the original API design and behavior as defined in that release, enabling transfer learning on top of a pre-trained MobileNet model for both classification and regression tasks.

Changes

- **Refactored model loading logic**: Instead of loading the full MobileNet and truncating inference at a specific layer (which varies between MobileNet v1 and v2), the new approach uses the `FeatureVector` variant of MobileNet directly. Its output is then fed into an MLP for downstream training, making the pipeline cleaner and more version-agnostic.

- **Fixed MLP input shape mismatch**: The previous implementation hardcoded the MLP input shape based on MobileNet v1 with `alpha=1`, without accounting for the `alpha` hyperparameter that scales filter sizes across the network. This fix dynamically resolves the correct input shape based on the chosen `alpha` value.

TODO

- [x] **Webcam Support** — Currently, the feature extractor only supports static image uploads. Add real-time webcam input so users can perform inference directly from their computer camera.

- [x] **Usage Examples** — Design example demos to demonstrate the feature extractor in practice. I will draw inspiration from @shiffman's [classification tutorial](https://www.youtube.com/watch?v=aKgq0m1YjvQ) and [regression tutorial](https://www.youtube.com/watch?v=eeO-rWYFuG0). Open to suggestions on example design!

- [x] **Model Save & Load** — Add the ability to save a trained model to disk and reload it in a later session, avoiding the need to retrain from scratch each time.


shiffman:
So excited about this! In addition to the examples I previously made regarding training your own classifier, some examples using cosine similarity could be good! (and perhaps we want to build into ml5 a cosine similarity utility function). Last year I made a demo that matches your webcam image to the most similar image. It has some unfriendly to beginners code, but could be a model for an ml5 version.

https://editor.p5js.org/ml_4_cc/sketches/CWE6Ox_jd

I am hoping to make some videos about this in the future.

JunhaoZhu0220:
Two newly implemented demos have been pushed to the `feature-extractor` branch:

1. **[featureExtractor-webcam-classifier](https://github.com/ml5js/ml5-next-gen/tree/feature-extractor/examples/featureExtractor-webcam-classifier)** — Classify between two categories (pens and phones) using the webcam.

2. **[featureExtractor-webcam-regressor](https://github.com/ml5js/ml5-next-gen/tree/feature-extractor/examples/featureExtractor-webcam-regressor)** — Control the size of a circle by moving your face closer to or farther from the camera.

Feel free to check it out!

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,69 @@
...
+    classifier.addImage("phone");
+    cupCount++;
+    console.log("Phone samples: " + cupCount);
+  });
+
+  trainButton = createButton("train");
+  trainButton.mousePressed(function () {
+    classifier.train().then(function () {
+      console.log("Starting classification...");
+      classifyVideo();
```
shiffman: Take a look at the new ml5.js `startClassify()` pattern in other examples. This eliminates the need for the "recursive" call to `classifyVideo()`!

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,69 @@
...
+  classifyVideo();
+}
+
+function setup() {
+  createCanvas(640, 540);
+  video = createCapture(VIDEO);
+  video.hide();
+  background(0);
+  // Initialize the feature extractor
+  featureExtractor = ml5.featureExtractor({ epochs: 100 }, modelReady);
```
shiffman: I think it makes more sense to have the `epochs` property (along with other parameters like `learningRate` be specified when calling the `train()` method on the `classifier` itself.

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,69 @@
...
+  // Create a new classifier using those features and with a video element
+  classifier = featureExtractor.classification(video, videoReady);
```
shiffman: This matches exactly how we implemented the API in ml5.js prior to 1.0. @gohai I think this could be up for discussion. Do we like this methodology of first loading the extractor and then "turning it into" a classifier? Another option would be to use `ml5.neuralNetwork()` in conjunction with the feature extractor. This might involve more code, but could be better from a pedagogical standpoint. Would love your thoughts!

JunhaoZhu0220: For the loading methodology, I have an idea of introducing a `task` property that accepts two options: `classification` and `regression`. With this approach, loading a classifier or regressor would look like:
```js
feClassifier = ml5.featureExtractor({ task: 'classification' }, modelReady);
feRegressor = ml5.featureExtractor({ task: 'regression' }, modelReady);
```

JunhaoZhu0220:
> So excited about this! In addition to the examples I previously made regarding training your own classifier, some examples using cosine similarity could be good! (and perhaps we want to build into ml5 a cosine similarity utility function). Last year I made a demo that matches your webcam image to the most similar image. It has some unfriendly to beginners code, but could be a model for an ml5 version.
> 
> https://editor.p5js.org/ml_4_cc/sketches/CWE6Ox_jd
> 
> I am hoping to make some videos about this in the future.

@shiffman @gohai For the cosine similarity utility, one demo that came up to my mind is handwritten digit recognition. Inspired by Yann Lecun et al.'s [MNIST database](https://en.wikipedia.org/wiki/MNIST_database), the feature extractor would first obtain feature embeddings for reference images of digits 0–9, then compute the embedding for a webcam frame where the user holds up a handwritten number on paper. The digit is identified by finding the reference embedding with the highest cosine similarity.

Please give me some feedbacks on how you feel about this idea!

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,69 @@
...
+  classifier.classify(gotResults);
+}
+
+function gotResults(results) {
+  label = results[0].label + " (" + nf(results[0].confidence, 0, 2) + ")";
+  classifyVideo();
+}
+
+function setup() {
+  createCanvas(640, 540);
```
gohai: (Nitpicky detail) Other examples are using a canvas resolution of 640 by 480 pixels with the webcam, and resizing its image to those dimensions.

examples/featureExtractor-webcam-classifier/sketch.js
gohai: The two-step process of creating an extractor and then using it to create a classifier does indeed seem quite confusing to me. I would prefer a single-line solution, similar to what @JunhaoZhu0220 is proposing.

examples/featureExtractor-webcam-classifier/index.html
```diff
@@ -0,0 +1,14 @@
...
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>ml5.js featureExtractor Example</title>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.11.10/p5.min.js"></script>
+    <script src="../../dist/ml5.js"></script>
+  </head>
+  <body>
+    <script src="sketch.js"></script>
+  </body>
+</html>
```
gohai: Tiny nitpick: Git ("diff" really) much prefers to have finishing newline character after the last line of text in a file

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,69 @@
...
+  phoneButton = createButton("phone");
+  phoneButton.mousePressed(function () {
+    classifier.addImage("phone");
+    cupCount++;
+    console.log("Phone samples: " + cupCount);
+  });
+
+  trainButton = createButton("train");
+  trainButton.mousePressed(function () {
+    classifier.train().then(function () {
```
gohai: At the moment, we aren't (yet) assuming that ml5 users are familiar with .then(), so it might be better to use a callback function instead.

With p5.js 2.0 adopting async/await, we might want to rethink this assumption at some point. But for now, the other examples are all using callbacks.

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,69 @@
...
+  video.hide();
+  background(0);
+  // Initialize the feature extractor
+  featureExtractor = ml5.featureExtractor({ epochs: 100 }, modelReady);
+  // Create a new classifier using those features and with a video element
+  classifier = featureExtractor.classification(video, videoReady);
+
+  penButton = createButton("pen");
+  penButton.mousePressed(function () {
+    classifier.addImage("pen");
```
gohai: Calling `addImage()` in the example before the model has finished loading throws the following exception:

```
Argument 'b' passed to 'sub' must be a Tensor or TensorLike, but got 'null'
    at convertToTensor (http://localhost:8080/dist/ml5.js:126317:15)
    at sub_ (http://localhost:8080/dist/ml5.js:115599:79)
    at sub__op (http://localhost:8080/dist/ml5.js:107479:28)
    at __webpack_modules__../node_modules/@tensorflow/tfjs-core/dist/public/chained_ops/sub.js.sub (http://localhost:8080/dist/ml5.js:124191:57)
    at http://localhost:8080/dist/ml5.js:291355:10
    at http://localhost:8080/dist/ml5.js:81133:22
    at Engine.scopedRun (http://localhost:8080/dist/ml5.js:81143:25)
    at Engine.tidy (http://localhost:8080/dist/ml5.js:81132:21)
    at Module.tidy (http://localhost:8080/dist/ml5.js:82406:56)
    at Mobilenet.normalizeAndResize (http://localhost:8080/dist/ml5.js:291346:58)
```

It might be easiest to thus make it so that the feature extractor gets loaded in `preload` (see e.g. the handPose example and library source code for how this is accomplished there).

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,69 @@
+let featureExtractor;
```
gohai: Thank you for this nice example inspired by @shiffman's video, @JunhaoZhu0220! 🤗

Since "pen" and "phone" are still fairly generic classes, and one might need to know that they aren't part of MobileNet to fully understand and apprechiate this functionality: wondering if making both buttons be `contenteditable` might be an idea worth trying out, in your opinion? (the user would be prompted to define their own classes, hits "done", and then continues in the way the example works currently)

examples/featureExtractor-webcam-classifier/sketch.js
gohai: Might be nice to support the training visualization via `debug: true` here as well. (This is available in the NeuralNetwork class.)

examples/featureExtractor-webcam-classifier/index.html
JunhaoZhu0220: Finishing newline characters have been added in [0b921f3](https://github.com/ml5js/ml5-next-gen/pull/301/commits/0b921f3e63781fa31ae4f8dcb488dccc413c622e).

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: For the model, the hyperparameters are able to be passed to .train() in [490ce7e](https://github.com/ml5js/ml5-next-gen/pull/301/commits/490ce7e0d052c709545c183829e162adedfd7352).

The training visualization and hyperparameter tuning when calling train() method in demos are available in [e119244](https://github.com/ml5js/ml5-next-gen/pull/301/commits/e119244b79a3f76b690169bc73019c4fb4ba53ea).

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: I have implemented my design in [e119244](https://github.com/ml5js/ml5-next-gen/pull/301/commits/e119244b79a3f76b690169bc73019c4fb4ba53ea), so now
```js
feClassifier = ml5.featureExtractor({ task: 'classification' }, modelReady);
feRegressor = ml5.featureExtractor({ task: 'regression' }, modelReady);
```
is the new way to initialize the model as shown in the two demos.

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: All .then() have been replaced with a callback function in [e119244](https://github.com/ml5js/ml5-next-gen/pull/301/commits/e119244b79a3f76b690169bc73019c4fb4ba53ea).

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: I have moved the model initialization within preload in [e119244](https://github.com/ml5js/ml5-next-gen/pull/301/commits/e119244b79a3f76b690169bc73019c4fb4ba53ea) to prevent this from happening.

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: The canvas resolutions in two demos have been all set to (640, 480) in [a1d3140](https://github.com/ml5js/ml5-next-gen/pull/301/commits/a1d31408236dd934399cf37a05bb34d0f9215bb2).

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: A new classifyStart() and predictStart() method has been implemented in [17afa84](https://github.com/ml5js/ml5-next-gen/pull/301/commits/17afa8496baeae541114f11c6e6b77dbc5c34d95).

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: @gohai I do think this is a perfect idea to offer the users the feasibilty of editing the classes on their own! The interface in [aa53f36](https://github.com/ml5js/ml5-next-gen/pull/301/commits/aa53f3607c19857fb5dee0e7844bc21812ea9731) looks like 
image
in which two classes will set defaultly to Class 1 and Class 2. If the user input customized class names in the text box, the names for the label will be changed accordingly.

gohai: This is already much more friendly @JunhaoZhu0220 ✨ 

What I meant with `contenteditable`, and which you might still want to try to see if you like it is the HTML attribute of the same name, which allows users to type in the button directly. (Might make the p5 code shorter, but I haven't tested myself.)

e.g.
Screenshot 2026-03-21 at 3 45 14 PM
Screenshot 2026-03-21 at 3 45 49 PM
Screenshot 2026-03-21 at 3 47 45 PM

(and if the user clicks "Done", then it might remain "Class #1" and "Class #2")

examples/featureExtractor-webcam-classifier/sketch.js
gohai: Wonderful! This works well.

examples/featureExtractor-webcam-classifier/index.html
gohai: Thanks for accommodating Git and my OCD 😅

examples/featureExtractor-webcam-classifier/sketch.js
gohai: I think that's great. Users now get the feature extractor right away, and the `task` property mirrors the NeuralNetwork class.

(Leaving for others to chime in)

examples/featureExtractor-webcam-classifier/sketch.js
gohai: Wonderful! Users who know `NeuralNetwork` will feel right at home...

examples/featureExtractor-webcam-regressor/sketch.js
```diff
@@ -0,0 +1,81 @@
+let feRegressor;
+let video;
+let slider;
+let addButton;
+let trainButton;
+let sampleCount = 0;
+let predictedValue = 0;
+
+function modelReady() {
```
gohai: Small nitpick: defining those function in code in the order that we expect them to be called (e.g. `preload` then `setup` etc) might make it _slightly_ easier to read the sketch top-to-bottom

examples/featureExtractor-webcam-regressor/sketch.js
```diff
@@ -0,0 +1,81 @@
...
+  console.log("Model is ready!");
+}
+
+function gotResults(results) {
+  predictedValue = results[0].value;
+}
+
+function preload() {
+  // Initialize the feature extractor for regression
+  feRegressor = ml5.featureExtractor({ task: 'regression', version: 2 }, modelReady);
```
gohai: Is the `version: 2` here necessary for it to function? (If yes, is there a drawback to changing our default from 1 to 2?)

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,75 @@
...
+  feClassifier = ml5.featureExtractor({ task: 'classification' }, modelReady);
+}
+
+function setup() {
+  createCanvas(640, 480);
+  video = createCapture(VIDEO);
+  video.hide();
+  background(0);
+  // Set the video as the input for the Classifier
+  feClassifier.video = video;
```
gohai: It is probably nicer to have a dedicated method to setting the input here (which might be a _video_, but possibly also an image, canvas...) 🤔 Something to think together with @shiffman at some convenient time.

For context: previously, the video input got passed as an argument to the constructor, but since the feature extractor now gets created in `preload()`, we typically don't have the video element yet at this point.
Our other models, such as bodyPose, work around this by taking `video` as an argument to e.g. `detectStart()`. We could do the same, and require video as an argument to both `addImage()` and `classifyStart()`. Or, pass the source to the feature extractor at one point in time only - similarly to how @JunhaoZhu0220 is doing here.

examples/featureExtractor-webcam-regressor/sketch.js
JunhaoZhu0220: The reason why I passed `version: 2` is that the regression task is different from what the mobilenet is originally designed for, so we might need to utilize a stronger (newer) model which can generalize the feature extraction better 🤗
If this may introduce some confusions to users, I can remove this passing and use the default `version: 1`, while indicating in the documentation that `version: 2` will bring a better performance.

gohai: Curious if changing the default to version 2 could be an option? Does this have downsides for the classification task perhaps?

JunhaoZhu0220: I believe version 2 also works perfectly for classification tasks. (will change the default version to 2 in next commit)

JunhaoZhu0220: Modified in [4f4c5f0](https://github.com/ml5js/ml5-next-gen/pull/301/commits/4f4c5f03b935bf100181fdec2bbc1f236717d80c) and [6b1dd36](https://github.com/ml5js/ml5-next-gen/pull/301/commits/6b1dd36f8b1bfccbd7f16b5bf856059489a239a1).

examples/featureExtractor-webcam-regressor/sketch.js
JunhaoZhu0220: The function orders have been changed in [bb16de0](https://github.com/ml5js/ml5-next-gen/pull/301/commits/bb16de04673ee42bfc9acce2936d42082f6ab870).

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: The contenteditable buttons have been introduced in the example in [78f2177](https://github.com/ml5js/ml5-next-gen/pull/301/commits/78f21776da487bf94a2207f3c6f78956420f2072).

JunhaoZhu0220:
In [6cf544e](https://github.com/ml5js/ml5-next-gen/pull/301/commits/6cf544e55c3d628c274864c13b66b8beb21784ae), I have introduced the function of  saving and loading model weights. 

The overall logic is:
`save()` bundles the MLP's architecture + weights + (task, labelIndex) metadata into `model.json` + `model.weights.bin` and triggers a browser download; `load()` accepts either a URL string or the two user-selected files (must be `.json` and `.weights.bin`), reconstructs the MLP, and restores task and label Index from the metadata, so the classifier predicts with its original class names, and the regressor still routes through the correct `predict()` branch.

@shiffman @gohai Feel free to check it out and test it!

examples/featureExtractor-webcam-classifier/sketch.js
shiffman: Thanks for this comment! We should explicitly reuqire the `video` element for each call to `addImage()` and `classifyStart()`. This then matches how the `ml5.imageClassifier()` works!

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,75 @@
...
+function modelReady() {
+  console.log("Model is ready!");
+}
+
+function gotResults(results) {
+  label = results[0].label + " (" + nf(results[0].confidence, 0, 2) + ")";
+}
+
+function preload() {
+  feClassifier = ml5.featureExtractor({ task: 'classification' }, modelReady);
```
shiffman: We can have a default model, but I prefer to explicitly pass in the model name for the examples. 

```js
  classifier = ml5.featureExtractor('MobileNet', { task: 'classification' }, modelReady);
```

Also, I find `feClassifier` to be an awkward name, I would probably go with `featureClassifier` or the plain `classifier` will do.

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,75 @@
...
+    count2++;
+    console.log(className + " samples: " + count2);
+  });
+
+  trainButton = createButton("train");
+  trainButton.mousePressed(function () {
+    feClassifier.train({ epochs: 100, debug: true }, function () {
+      console.log("Starting classification...");
+      feClassifier.classifyStart(gotResults);
+    });
```
shiffman: I might suggest a "while training" callback that includes the loss value. And to be more beginner friendly, lI would suggest using a named / defined function. Something like:

```js
trainButton.mousePressed(startTraining);

function startTraining() {
  classifier.train({ epochs: 100, debug: true }, whileTraining);
}

function whileTraining(loss, epoch, status) {
  if (status == 'done') {
    classifier.classifyStart(video, gotResults);
  }
}
```

Not sure if `'done'` is the best value for `status`. Curious for @gohai's feedback on this!

examples/featureExtractor-webcam-regressor/sketch.js
shiffman: Noting we should explicity add the model argument: `'MobileNet'`

examples/featureExtractor-webcam-regressor/sketch.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+function setup() {
+  createCanvas(640, 480);
+  video = createCapture(VIDEO);
+  video.hide();
+  background(0);
+  // Set the video as the input for the Regressor
+  feRegressor.video = video;
```
shiffman: Same comment here, `video` should be passed in when calling `startPredict()` or `addImage()`.

shiffman: Also, I wonder if the variable name `predictor` to go with the `predict()` function would be best here.

examples/featureExtractor-webcam-regressor/sketch.js
```diff
@@ -0,0 +1,81 @@
...
+
+  // Draw the video (flipped horizontally)
+  push();
+  translate(640, 0);
+  scale(-1, 1);
+  image(video, 0, 0, 640, 450);
+  pop();
+
+  // Use predicted value if trained, otherwise follow the slider
+  let currentValue = feRegressor.isTrained ? predictedValue : slider.value();
```
shiffman: The ternary operator is likely unfamiliar to beginner p5.js or ml5.js users.

shiffman (commented):
Let another round of comments, hope these are helpful!

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: The plain `classifier` seems to be more intuitive. (and also `predictor` in regression example) I will change them accordingly.

Regarding whether we should let users pass in the model name, my concern is that the feature extractor's codebase only supports MobileNet. It isn't set up to load weights from other models.
If we expose a model parameter in the API, I worry it could be misleading: users would reasonably assume they can pass in other model names.

@gohai @shiffman Please tell me about your ideas on this!

examples/featureExtractor-webcam-regressor/sketch.js
JunhaoZhu0220: The operator has been replaced with an if statement in [2c0da94](https://github.com/ml5js/ml5-next-gen/pull/301/commits/2c0da945f390ec902d07124a06238144a8065d7c).

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: The `video` element can be passed to all related functions in [aca085f](https://github.com/ml5js/ml5-next-gen/pull/301/commits/aca085f0b02cb4e9bff02a01bb7bc16c3960ba6c), also the naming has been changed for clarity in [2c0da94](https://github.com/ml5js/ml5-next-gen/pull/301/commits/2c0da945f390ec902d07124a06238144a8065d7c).

examples/featureExtractor-webcam-regressor/sketch.js
JunhaoZhu0220: Solved in [aca085f](https://github.com/ml5js/ml5-next-gen/pull/301/commits/aca085f0b02cb4e9bff02a01bb7bc16c3960ba6c) and [2c0da94](https://github.com/ml5js/ml5-next-gen/pull/301/commits/2c0da945f390ec902d07124a06238144a8065d7c).

examples/featureExtractor-webcam-classifier/sketch.js
gohai: Seconding the suggestion to add the model name as the first parameter. Even if it isn't "doing" anything presently: (a) it's future-proofing the API for possible support of other models, and perhaps more importantly (b) it makes it explicit to the user which pre-trained model this is using under the hood.

For now, the library could simply return an error if the first argument isn't "MobileNet" - perhaps with a list (of one) of possible models in the console.

examples/featureExtractor-webcam-classifier/sketch.js
gohai: Agree - anonymous functions as parameters might be a bit confusing to beginners.

@shiffman Since the callback for `NeuralNetwork.train()` signals completion, I would naively lean towards keeping the way the operate in sync. 🤔 (The use case you are imagining is custom status displays? Not programmatically cancelling out of the training based on loss, right?)

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+function setup() {
+  let cnv = createCanvas(640, 480);
+  cnv.parent("canvasContainer");
+  // Show the name controls now that the canvas exists
+  document.getElementById("nameControls").style.display = "";
+  video = createCapture(VIDEO);
+  video.hide();
+  background(0);
```
gohai, line 23: Not needed?

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,89 @@
...
+
+  // Grab the editable name buttons + Done button defined in index.html
+  nameButton1 = select("#nameButton1");
+  nameButton2 = select("#nameButton2");
+  doneButton = select("#doneButton");
+
+  doneButton.mousePressed(function () {
+    // Get class names from the editable buttons, or use defaults if unchanged
+    let text1 = nameButton1.elt.textContent.trim();
```
gohai, line 32: `nameButton.value()` to be more p5-idiomatic?

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,89 @@
...
+  });
+}
+
+function draw() {
+  background(0);
+
+  // Draw the video (flipped horizontally)
+  push();
+  translate(640, 0);
+  scale(-1, 1);
```
gohai, line 73: Is there a reason to flip the image? (And if yes: is there a reason _not_ to do it with `flipped: true` in `createCapture`? E.g. does the performance of the model suffer?)

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,89 @@
...
+  image(video, 0, 0, 640, 450);
+  pop();
+
+  // Draw the label
+  fill(255);
+  textSize(16);
+  text(label, 10, height - 10);
+}
+
+function modelReady() {
```
gohai, line 83: Wondering if we want the `modelReady` in the example, now that the feature extractor is being created in `preload` 🤔

examples/featureExtractor-webcam-classifier/sketch.js
```diff
@@ -0,0 +1,89 @@
...
+  // Grab the editable name buttons + Done button defined in index.html
+  nameButton1 = select("#nameButton1");
+  nameButton2 = select("#nameButton2");
+  doneButton = select("#doneButton");
+
+  doneButton.mousePressed(function () {
+    // Get class names from the editable buttons, or use defaults if unchanged
+    let text1 = nameButton1.elt.textContent.trim();
+    let text2 = nameButton2.elt.textContent.trim();
+    class1 = (text1 && text1 !== "Click to edit Class #1") ? text1 : "Class #1";
```
gohai, line 34: Another ternary operator here: wondering if this can be written simpler - e.g. with `replace()`?

examples/featureExtractor-webcam-classifier/sketch.js
shiffman: I'd also be ok with more of a "warning" style console log explaining the MobileNet model is being loaded as default.

examples/featureExtractor-webcam-classifier/sketch.js
shiffman: @gohai I believe it actually optionally supports one or two callbacks! I'm not sure if we have any examples that make use of `whileTraining()` but I've demonstrated it in class before. Take a look and let me know what you think.

https://github.com/ml5js/ml5-next-gen/blob/main/src/NeuralNetwork/index.js#L481

gohai: Oh wow - I never knew this existed. Thank you for excavating this 🙌

Should we follow the same arguments then? e.g. the first function is understood to be the `finishedTrainingCb`, and a second one is the `whileTrainingCb`

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: For here, I found a mismatch that if I created `contenteditable` buttons from html, I might not be able to fetch the text element inside the buttons with `.value()` function in javascript.
Alternatively, I have come up with a more intuitive p5-idiomatic method. The two inputs for the two classes is now initialized with
```
nameInput1 = createInput(class1);
nameInput2 = createInput(class2);
```
and it will look like this:
image
The fetch of the text elements will be done with:
```
class1 = nameInput1.value().trim() || "Class #1";
class2 = nameInput2.value().trim() || "Class #2";
```
So that if no edit is done, the two classes will stay at `Class #1` and `Class #2`, i.e.
image
or otherwise be the user defined class names, i.e.
first:
image
after clicking `Done`:
image

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: Indeed `modelReady` is no longer necessary! I will remove it.

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: I have added the model name as the first parameter, and utilized the error messages warning in `src/utils/handleOptions.js`.
Currently if the users used a model name other than `MobileNet` (e.g. `ResNet`), a console log warning will be shown to them:
```
🟪ml5.js warns: The modelName parameter for featureExtractor has to be set to 'MobileNet', but it is being set to 'ResNet' instead.

ml5.js is using default model 'MobileNet'.
```
and they continue with the default MobileNet models and configs.

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: The reason why I flipped the image is that I hope the video shown on the canvas can mirror what the users are doing.

Thank you for mentioning `flipped: true` in `createCapture`! This seems to be a better solution than directly flipping the canvas.

examples/featureExtractor-webcam-regressor/sketch.js
JunhaoZhu0220: Noted!

examples/featureExtractor-webcam-classifier/sketch.js
JunhaoZhu0220: Currently in `setup()`, I only utilized
```
createCanvas(640, 480);
video = createCapture(VIDEO, { flipped: true });
video.hide();
background(0);
```
to set the canvas together with the webcam streaming, and hide the actual webcam image.

gohai: Looks like my context was a bit off - this was meant specifically for the `background()` line

JunhaoZhu0220: Yes, this is not needed 😂 Remove it immediately!

gohai:
I believe this PR is ready to be merged! (Ideally via squash and merge.)

Admittedly, testing on the regression example has been somewhat light, but I believe any necessary tweaks can be addressed in follow-up PRs and commits ahead of a pre-fall-semester release. (We may also want to clean up examples and documentation to fully align with p5 2.0 at that point anyway, since the p5 Web Editor is scheduled to switch over then as well.)

I’d be happy to take this on (and will CC @JunhaoZhu0220 for review). He already laid the foundation for the documentation in https://github.com/ml5js/ml5-website-v02-docsify/pull/227 🥳

@JunhaoZhu0220 Don't forget to add yourself to the `all-contributors` section in the `README.md`!

JunhaoZhu0220:
The feature-extractor branch has been squashed and merged to the main branch.
Thanks @shiffman and @gohai for all the comments and feedbacks! Feel free to tag me if there is any issue.

I will keep working on the documentations, hopefully (😭) I can finish a first draft by early June.

PR #304, opened 2026-03-29
feat(models): add offline model caching via IndexedDB

alanvww:
## Summary
This pull request adds support for offline model caching using the browser's IndexedDB. By passing `{ cache: true }` when initializing models like `handPose`, `bodyPose`, and `faceMesh`, the models are transparently downloaded and stored locally. On subsequent page loads, the models are served directly from the IndexedDB cache, making the library fully functional without an internet connection (ideal for installations and exhibitions).

## Changes
*   **Model Caching:** Added `CachingIOHandler` in `src/utils/modelCache.js` to intercept TensorFlow.js model requests and store them in IndexedDB.
*   **Model Registry:** Added `src/utils/modelRegistry.js` to resolve dynamic model URLs for caching.
*   **Global Utilities:** Introduced `ml5.isCached()`, `ml5.cacheModel()`, and `ml5.clearCache()` to manually manage the IndexedDB state.
*   **Robust Initialization:** Added `await this.ready;` across `detect()` and `detectLoop()` methods in all major models to prevent synchronous crashes when `this.model` is `null` during background initialization.
*   **Examples:** Added one example of handpose offline available at `examples/handPose-keypoints-offline` to showcase how to preload and use models in an offline environment, keeping the UI minimal and consistent with other examples.

### DIFF
--- a/examples/handPose-keypoints-offline/index.html
+++ b/examples/handPose-keypoints-offline/index.html
@@ -1,0 +1,50 @@
+<!--
+  👋 Hello! This is an ml5.js example made and shared with ❤️.
+  Learn more about the ml5.js project: https://ml5js.org/
+  ml5.js license and Code of Conduct: https://github.com/ml5js/ml5-next-gen/blob/main/LICENSE.md
+
+  This example demonstrates offline-capable hand tracking using ml5.handPose
+  with the { cache: true } option. On first load the model is downloaded from
+  the network and saved to IndexedDB. On every subsequent load the model is
+  served from the local cache — no internet connection required.
+-->
+
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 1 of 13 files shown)


shiffman:
Thank you for posting this @alanvww!

I love the idea of using the browser cache / localStorage! I think this could be quite convenient for many use cases. I have a few questions about it.

1. Would there be an issue with a machine that is set up to boot into an ml5.js sketch? Is the cache retained if the browser quits and restarts?
2. Is downloading the model files directly and placing them in a relative path to the sketch an option as well? We do this with `ml5.neuralNetwork()` and I wonder if this pattern would make sense to follow.

```js
  let modelDetails = {
    model: "model/model.json",
    metadata: "model/model_meta.json",
    weights: "model/model.weights.bin",
  };

  classifier.load(modelDetails, modelLoaded);
```

Thank you for the excellent ml5.js keypoints example! I might put this more in the category of "demo" since it uses a lot of code that will be unfamiliar to beginner users of p5.js and ml5.js (ternary operators, native DOM functions, try/catch, etc.) I might suggest making a drop dead simple version of it that takes the existing examples and only adds the automatic loading/caching. Instead of DOM elements you can console log the sketch's status.

PR #306, opened 2026-05-31
Fix iOS WebGPU keypoint misalignment

### LINKED ISSUE #302: facemesh keypoints NOT aligned with video on mobile device
### Issue Type

Runtime Error

### ml5.js Version

v1.3.1

### Development Environment

local setup, p5js editor

### Browser & Operating System

Safari 26.3, Chrome 145 on iOS 26.3,

### Issue Description

- run face mesh example code on iOS mobile device in Safari or Chrome
- [faceMesh-keypoints](https://editor.p5js.org/ml5/sketches/lCurUW1TT)
- OR: run facemesh example from [facemesh reference](https://docs.ml5js.org/#/reference/facemesh)
- expected behavior: facemesh keypoints aligned with video.
- actual behavior: facemesh keypoints NOT aligned with video

- Not an issue on desktop with ml5@1.3.1
- Not an issue on mobile with ml5@1.2.0 (although you get error message "Initialization of backend webgpu failed")

Image

### Minimal Reproducible Example

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

alanvww:
Fixes #302 — on iOS (iOS 26+, Safari/Chrome), faceMesh and handPose keypoints show up rotated ~90° and out of line with the video. This worked in v1.2.0.

## Why it happens

It's a TF.js backend issue, not ml5's keypoint code (that code didn't change between 1.2.0 and 1.3.1).

ml5 bundles the WebGPU backend, which TF.js picks over WebGL when it's available. On iOS, WebGPU reads video frames with the wrong orientation, so the keypoints come out rotated. In 1.2.0 WebGPU failed to start on iOS and fell back to WebGL, so it looked fine. iOS 26 + the newer TF.js now run WebGPU successfully, which is when the bug shows up.

## The fix

Turn off the `WEBGPU_IMPORT_EXTERNAL_TEXTURE` flag so WebGPU copies video frames instead. Keypoints line up again, and we keep using WebGPU.

- `src/utils/configureBackend.js` (new) — sets the flag
- `src/index.js` — calls it once on load
- `tests/unit/configureBackend.test.js` (new)

Note: this adds a small per-frame copy on all WebGPU devices, including desktop where there's no bug. WebGL is unaffected, and `ml5.setBackend(...)` still works.

## Testing

- iOS 26 (Safari + Chrome): faceMesh and handPose keypoints now line up. 
- `npm test` passes, including the new test.

### Example sketches
- OLD library: https://editor.p5js.org/alanvww/sketches/YRV8Fke5v
- OLD library with forced WebGL: https://editor.p5js.org/alanvww/sketches/KHrvzVgo2
- FIXED library: https://editor.p5js.org/alanvww/sketches/32YR4JF5H

## Update

- Reported upstream as tensorflow/tfjs#8733 (may ultimately be a WebKit `importExternalTexture` bug). The workaround can be removed once that's fixed; the code comment links it.
- The console notice now only logs on iOS/iPadOS with WebGPU available — the platforms actually affected — instead of every WebGPU-capable browser.
- Verified on device: faceMesh and handPose on iOS 26. bodyPose uses the same video input path so it should be equally fixed, but hasn't been device-verified yet — testing welcome.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,6 +8,7 @@ import soundClassifier from "./SoundClassifier";
 import objectDetection from "./ObjectDetection";
 import featureExtractor from "./FeatureExtractor";
 import setBackend from "./utils/setBackend";
+import configureBackend from "./utils/configureBackend";
 import bodySegmentation from "./BodySegmentation";
 import depthEstimation from "./DepthEstimation";
 import communityStatement from "./utils/communityStatement";
@@ -47,6 +48,9 @@ p5Utils.setupP5Integration(
   Object.keys(withPreload),
   Object.keys(withoutAsync)
 );
+
+// Apply ml5's default TensorFlow.js backend configuration (see issue #302).
+configureBackend();
 
 communityStatement();
 
--- a/src/utils/configureBackend.js
+++ b/src/utils/configureBackend.js
@@ -15,6 +15,9 @@ const WEBGPU_VIDEO_TEXTURE_FLAG = "WEBGPU_IMPORT_EXTERNAL_TEXTURE";
  * The flag is set globally and unconditionally (it only affects the WebGPU
  * backend, so it is harmless when WebGL is active). Safe to call more than once.
  *
+ * Reported upstream as https://github.com/tensorflow/tfjs/issues/8733 — the
+ * workaround can be removed once tf.js (or WebKit) fixes the import path.
+ *
  * @return {void}
  */
 export default function configureBackend() {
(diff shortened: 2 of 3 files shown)


gohai:
Many thanks @alanvww for getting to the bottom of this issue, and offering this fix 🤗

Since you already have the logic to determine if the user is on an iOS device: wondering if it'd make sense to _only_ set the flag to force texture copies under those conditions? (I am guessing that this incurs some performance penalty that might be unnecessary on other platforms?)

alanvww:
> Many thanks @alanvww for getting to the bottom of this issue, and offering this fix 🤗
> 
> Since you already have the logic to determine if the user is on an iOS device: wondering if it'd make sense to _only_ set the flag to force texture copies under those conditions? (I am guessing that this incurs some performance penalty that might be unnecessary on other platforms?)

aw good call! Yes, that actually make sense, will add that soon!

shiffman (commented):
I had the same thought at @gohai regarding only applying the change if iOS is detected. Otherwise I think the PR looks great! I'll be glad to merge when ready. We may also want to do a patch release following the merge. (cc @colemanliyah, if you want to learn about the release process we could schedule a time to doo this over Zoom!)

colemanliyah:
> I had the same thought at @gohai regarding only applying the change if iOS is detected. Otherwise I think the PR looks great! I'll be glad to merge when ready. We may also want to do a patch release following the merge. (cc @colemanliyah, if you want to learn about the release process we could schedule a time to doo this over Zoom!)

@shiffman I'd love to learn about the release process!

alanvww:
@gohai @shiffman Thanks both! Pushed an update — the flag is now only set when iOS/iPadOS is detected, so WebGPU keeps its zero-copy path everywhere else.

shiffman:
Thank you @alanvww! Merging this and it will be in the next release.

PR #307, opened 2026-06-05
Fix some minor issues with the feature extractor.

JunhaoZhu0220:
This PR builds on the previously [merged PR](https://github.com/ml5js/ml5-next-gen/pull/301). While writing the Methods section of the [documentation](https://github.com/ml5js/ml5-website-v02-docsify/pull/227), I noticed that some return values were redundant and could be removed, and a few variable names were unclear, which made them hard to explain in the docs. I therefore did another round of code review and made changes to address these issues.


gohai:
Many thanks @JunhaoZhu0220, for caring about this feature 🤗 Let me go ahead and merge the commit right away, since it only touches the implementation. I am planning on doing some overall testing soon

gohai:
Done!

Small process suggestion: after we've squash-merged a branch into `main`, it's usually best to start any follow-up work on a new branch created from the updated `main`.

What happened here is that your `feature-extractor` branch still contained the original commit history that was collapsed when we squash-merged it. Since that history no longer exists on main, Git sees the branches as having diverged, which is why there are conflicts.

FWIW: Quick way to resolve this is checking out `main` and git cherry-pick`ing just your latest fix commit.

PR #308, opened 2026-06-26
Remove p5.js 1.x examples

gohai:
This moves the p5.js 2.x examples to the spot of their p5.js 1.x counterparts. (Uploading the examples to the web editor after the next release should automatically update those to the p5.js 2.x syntax while retaining the URL.)

This has some simplifications to the scripts that haven't been tested yet.

Also some tweaks to the FeatureExtractor (which is currently also not yet on the Web Editor).


ksen0:
Hi @gohai , thanks for the very quick response! Do I understand correctly, that this would update all ml5 examples on the web editor? The reason I ask is that after [the website switch](https://github.com/processing/p5.js-website/issues/1426) the v1 website will still be up at v1.p5js.org. The [tutorial](https://p5js.org/tutorials/speak-with-your-hands/) would still be there, but if it links to version 2 live examples, that might be good to add a note or something. 

Alternatively you could embed the examples in their entirety on the tutorial page. (Some recent updates to parameterizing [the code embed blocks](https://github.com/processing/p5.js-website/pull/1469) more; it is also possible to [load additional scripts into editable sketches](https://github.com/processing/p5.js-website/blob/2.0/src/content/tutorials/en/typography-2.0.mdx?plain=1#L115).) So if the code is different for v1, it is possible to create a self-contained archived tutorial variant, if that's preferred.

gohai:
@ksen0 Indeed, as this stands now, this would switch over all the examples once we run the script (typically after releasing a version). I had a closer look at the [tutorial](https://p5js.org/tutorials/speak-with-your-hands/) you linked: turns out that this is still making use of the pre-1.0 version of ml5, with the web-editor links pointing to the tutorial authors' personal sketches. Ideally, this would get updated to ml5 1.x as well (the API changed, so this might entail some textual changes too). I'll make a note about this in the issue as well.

gohai:
@shiffman Please consider merging this either before or after the patch release you mentioned! (I think this should be done alongside updating our documentation to assume p5.js 2.0. Announcing the feature detector could happen at the same time as well. I still need to review the user-facing examples a bit more for that, though...)

shiffman:
Thank you for the reminder @gohai! I've opened https://github.com/ml5js/ml5-website-v02-docsify/issues/236 to track the move to p5.js 2.

shiffman:
Now that the docs are updated for p5.js 2.0 I'm merging this branch!

PR #309, opened 2026-08-06
FeatureExtractor: Tweaks to examples

gohai:
Hi @JunhaoZhu0220 (& @shiffman)

Here a round of stylistic tweaks to your fantastic examples (some details in the first commit). Hope this makes sense to you, and if I accidentally broke anything 😅

### DIFF
--- a/examples/featureExtractor-webcam-regressor/index.html
+++ b/examples/featureExtractor-webcam-regressor/index.html
@@ -4,19 +4,19 @@
     <meta charset="UTF-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>ml5.js featureExtractor Regression - Distance Control</title>
+    <title>ml5.js featureExtractor Regression Example - Distance Control</title>
     <script src="https://cdn.jsdelivr.net/npm/p5@2.0.5/lib/p5.js"></script>
     <script src="../../dist/ml5.js"></script>
   </head>
   <body>
     <h2>Feature Extractor: Regression - Distance Control</h2>
-    <p>Control the size of a circle by moving closer to or farther from the camera.</p>
+    <p>Control the size of a circle by moving closer or further away from the camera.</p>
     <ol>
       <li><strong>Lean back</strong> from the camera, set the slider to <strong>0</strong>, and click "Add Sample" several times.</li>
       <li><strong>Lean in close</strong> to the camera, set the slider to <strong>1</strong>, and click "Add Sample" several times.</li>
       <li>Optionally, add samples at <strong>middle distances</strong> with the slider around <strong>0.5</strong>.</li>
       <li>Click <strong>"Train"</strong> and wait for training to complete.</li>
-      <li>Move closer or farther from the camera — the yellow circle will grow and shrink in real time!</li>
+      <li>Move closer or further away from the camera - the yellow circle will grow and shrink in real time!</li>
     </ol>
     <script src="sketch.js"></script>
   </body>
--- a/examples/featureExtractor-webcam-classifier/sketch.js
+++ b/examples/featureExtractor-webcam-classifier/sketch.js
@@ -1,81 +1,83 @@
 let classifier;
 let video;
-let label = "";
-let nameInput1, nameInput2;
+
+let classButtons = [];
 let doneButton;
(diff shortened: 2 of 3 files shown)


JunhaoZhu0220:
@gohai The tweaks look great to me! For the contenteditable button in the classifier, one small concern about the rename flow:

Clicking on "Class 1 (click to rename)" places the cursor inside the button but keeps the full text, so the user has to select and delete everything before typing a new name. The code then recovers the default name by suffix matching:
```
let label = classButtons[i].html();
if (label.endsWith(" (click to rename)")) {
  label = label.slice(0, -18);
}
```

This feels a bit fragile:

- The -18 is the hardcoded length of " (click to rename)" — if the hint text ever changes, the check no longer works.
- The `endsWith` check only holds if the user leaves the text fully untouched. Any partial edit, or pressing Enter (which inserts a `<br>` after the suffix) will cause the code to skip the if statement.

Would it make sense to clear the text on first focus instead, and fall back to a default name? Something like:
```
for (let i = 0; i < 2; i++) {
  let button = createButton("Class #" + (i + 1) + " (click to rename)");
  button.attribute("contenteditable", "true");
  button.elt.addEventListener("focus", function () {
    this.innerText = "";
  }, { once: true });
  classButtons.push(button);
}
```

and in startSampling():
```
let label = classButtons[i].html().trim();
if (label === "" || label.endsWith("(click to rename)")) {
  label = "Class #" + (i + 1);
}
```
Not sure about whether this can be a good fix (as it is way more complicated than the current code), feel free to exchange ideas with me!

gohai:
@JunhaoZhu0220 How do you think about cb5caef2394b03f32f3d45265559ecfc3b526cff?

Reaching into `.elt` and having an anonymous function inline seemed a bit too complicated imho

JunhaoZhu0220:
https://github.com/ml5js/ml5-next-gen/commit/cb5caef2394b03f32f3d45265559ecfc3b526cff This commit looks great and very intuitive! Love this.

examples/featureExtractor-webcam-regressor/index.html
```diff
@@ -4,19 +4,19 @@
...
     <h2>Feature Extractor: Regression - Distance Control</h2>
-    <p>Control the size of a circle by moving closer to or farther from the camera.</p>
+    <p>Control the size of a circle by moving closer or further away from the camera.</p>
     <ol>
       <li><strong>Lean back</strong> from the camera, set the slider to <strong>0</strong>, and click "Add Sample" several times.</li>
       <li><strong>Lean in close</strong> to the camera, set the slider to <strong>1</strong>, and click "Add Sample" several times.</li>
       <li>Optionally, add samples at <strong>middle distances</strong> with the slider around <strong>0.5</strong>.</li>
       <li>Click <strong>"Train"</strong> and wait for training to complete.</li>
-      <li>Move closer or farther from the camera — the yellow circle will grow and shrink in real time!</li>
+      <li>Move closer or further away from the camera - the yellow circle will grow and shrink in real time!</li>
```
shiffman, line 19: This is great but i wonder if we should add one extra sentence along the lines of:

> This is just one way to train the model. You could also try mapping the slider value to horizontal or vertical movements, or experiment with various objects, sizes, and positions in the camera view.

shiffman (commented):
Love these examples, thank you @JunhaoZhu0220 and @gohai! I'll make sure to merge tomorrow before we do the release, I left one small comment if you want to address now or we can at a later time.

examples/featureExtractor-webcam-regressor/index.html
gohai: No need to include this, but - generally - I'd be curious how well the Feature Extractor would would handle different _states_ of related objects: e.g. bananas at different stages of ripeness

JunhaoZhu0220: Honestly I am not sure about how the feature extractor will perform in such an abstract case 😂 From my perspective, when given images of bananas at different ripeness stages, the MobileNet model would likely rely on color cues (green for unripe, yellow for ripe) as well as surface patterns (e.g., black spots on the peel) to estimate the degree of ripeness. However, since the prediction depends on multiple visual factors, the overall accuracy may not be very high.

I have found a [related dataset of various banana ripenesses](https://www.kaggle.com/datasets/lucianon/banana-ripeness-dataset) and can definitely give it a try!

shiffman:
Merging for the release!

PR #313, opened 2026-08-07
Upload Examples Fix

shiffman:
During the release, @colemanliyah and I noticed the upload script was broken. I only had a few minutes to debug, but this seemed to fix it. @pearmini maybe you can take a look when you have a chance?

### DIFF
--- a/scripts/updateExamplesREADME.js
+++ b/scripts/updateExamplesREADME.js
@@ -157,7 +157,7 @@ async function main() {
   console.log("\n📋 Summary:");
   console.log(`- p5.js 2.0 examples: ${exampleDirs.length}`);
   console.log(
-    `- Total sketches found on p5.js editor: ${sketchesRes.data.length}`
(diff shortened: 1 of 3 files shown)


gohai:
I believe the output of the `/editor/(username)/projects` endpoint must have changed, and is now wrapped inside an object with a `projects` key. LGTM @shiffman 🙇‍♂️

Screenshot 2026-08-08 at 6 09 08 AM

scripts/updateExamplesREADME.js
pearmini, line 161: Optionally fix: After the API change that prints undefined; it should be sketchesRes.data.projects.length. Doesn’t break upload/README generation!

pearmini (dismissed):
LGTM! safe to merge for the reported issue; optionally fix that summary count in a follow-up!

shiffman:
Not sure why it says I "dismissed" your review @pearmini, I made the change and clicked "resolve comment"! The opposite of dismissal! Anyway, this is all set now, going to go ahead and merge!

PR #314, opened 2026-08-07
Update README example sketches

shiffman:
Now includes feature extractor examples. This is the result of running the fixed #313 code. Not sure why some of the neural network examples got capitalized and then listed first. Perhaps I should manually change these back to lowercase?


gohai:
This looks good to me! (Glad it's working 🥳)

@shiffman My hunch is that those folders could still be capitalized on the machine where the script ran. We might have changed their names in a later commit to use the capitalization `neuroEvolution`, but because macOS file systems aren't case-sensitive, maybe the change didn't propagate back? Not sure, though!

shiffman:
Thanks @gohai, indeed this was the case, my local folders were capitalized. I fixed that and ran the script again. Closing this PR as the updated README is now in #313

PR #316, opened 2026-08-16
ImageClassifier: Fix Teachable Machine example

gohai:
These examples must have been inadvertently broken during the p5 2.x compatibility changes.

(Spotted while reviewing @colemanliyah's branch)


shiffman:
Thanks for catching this!

